PR #65, opened 2024-06-28
Use JSON Lines

edsu:
This is an experiment in using JSON Lines for our harvested JSON data instead of CSV.

This commit moves our sul_pub, dimensions and openalex harvesters over
to writing records as JSON Lines (jsonl). This means we will preserve
the dictionary and list data structures we received from the APIs and
will be able to use them for querying later, e.g. ORCID IDs embedded in
author objects.

Reading in JSON-L can be achieved with Pandas like:

```python
pandas.read_json('data.jsonl', orient='records', lines=True)
```

and Polars:

```python
polars.read_ndjson('data.jsonl')
```

or lazily:

```python
polars.scan_ndjson('data.jsonl')
```

### DIFF
--- a/rialto_airflow/dags/harvest.py
+++ b/rialto_airflow/dags/harvest.py
@@ -5,8 +5,7 @@ from airflow.models import Variable
 from airflow.decorators import dag, task
 
 from rialto_airflow.utils import create_snapshot_dir, rialto_authors_file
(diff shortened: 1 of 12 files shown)


edsu:
This was a useful experiment to see how the structure of the harvested data could be preserved. But I don't think it's strictly necessary. Perhaps it's something to keep in mind for the future though?

PR #73, opened 2024-07-03
Normalize DOIs in doi_sunet and merge_pubs tasks

### LINKED ISSUE #72: Normalize DOIs from all sources
In order to create a useful doi -> sunet dictionary, we need to normalize DOIs from sul_pub, OpenAlex, and Dimensions. This could be done within the `doi_sunet` task. 

The task should:
* downcase all DOIs since they can be capitalized differently across sources (e.g. `10.1103/PhysRevLett.96.07390` vs `10.1103/physrevlett.96.07390`. Confirm that Dimensions and sul_pub will find the DOIs when looked up in the downcased format. (Already confirmed that OpenAlex lookups are case-insensitive.)
* Remove leading/trailing whitespace
* Remove leading `DOI: ` or `doi: `
* Remove `http://dx.doi.org/` 

There are also DOIs in the sul_pub data that have internal spaces. I'm not sure if we can reliably clean these up. It might be worth correcting the pub hash in sul_pub. Examples: 
`113101 10.1103/PhysRevB.76.113101`
`10.1007/978-3-319-46759-7 23`

lwrubel:
Resolves #72 to make DOIs consistent across sources for accurate doi->sunet mapping and publication merging. 

There are still some sul_pub DOIs that will be malformatted for matching because of the variability of that data source. Just hoping that they do not cause any problems in lookups downstream. Will test by running the branch on prod with the full data. 

Ran DAG locally with a small dev limit and that succeeded.

### DIFF
--- a/rialto_airflow/harvest/doi_sunet.py
+++ b/rialto_airflow/harvest/doi_sunet.py
@@ -5,6 +5,8 @@ from collections import defaultdict
 # TODO: use polars instead?
 import pandas as pd
 
+from rialto_airflow.utils import normalize_doi
+
 
 def create_doi_sunet_pickle(
     dimensions: str, openalex: str, sul_pub_csv: str, authors_csv: str, output_path
@@ -42,6 +44,7 @@ def doi_sunetids(pickle_file: str, orcid_sunet: dict) -> dict:
 
     mapping = {}
     for doi, orcids in doi_orcids.items():
+        doi = normalize_doi(doi)
         mapping[doi] = [orcid_sunet[orcid] for orcid in orcids]
 
     return mapping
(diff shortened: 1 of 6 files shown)


lwrubel:
The normalization in this PR will remove colons from `doi: ` or `DOI: ` in a doi value. If there are other colons in a DOI, then we will get the same error from OpenAlex, 
```
"error: "Invalid query parameters error.",
message: "It looks like you're trying to do an OR query between filters and it's not supported. 
You can do this: institutions.country_code:fr|en, but not this: institutions.country_code:gb|host_venue.issn:0957-1558. 
Problem value: 10.1103/PhysRevB.75.075422|doi: 10.1103/PhysRevB.84.075477; 4599"
```

It will be helpful to see what those values might be (if they exist).

jacobthill (approved):

PR #78, opened 2024-07-09
Handle single DOI invalid in OpenAlex

lwrubel:
OpenAlex handles DOIs with commas in a way that wasn't expected and may return an empty list before erroring on the second half.

### DIFF
--- a/rialto_airflow/harvest/openalex.py
+++ b/rialto_airflow/harvest/openalex.py
@@ -100,7 +100,8 @@ def publications_from_dois(dois: list):
                     pubs = Works().filter(doi=doi).get()
                     if len(pubs) > 1:
                         logging.warn(f"Found multiple publications for DOI {doi}")
-                    yield normalize_publication(pubs[0])
+                    if len(pubs) > 0:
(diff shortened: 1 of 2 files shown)


rialto_airflow/harvest/openalex.py
```diff
@@ -100,7 +100,8 @@ def publications_from_dois(dois: list):
                     pubs = Works().filter(doi=doi).get()
                     if len(pubs) > 1:
                         logging.warn(f"Found multiple publications for DOI {doi}")
-                    yield normalize_publication(pubs[0])
+                    if pubs:
```
edsu, line 103: Should this be an `else`?

edsu (approved):

rialto_airflow/harvest/openalex.py
edsu: Actually I understand now. Maybe it should be `if len(pubs) > 0` ?

lwrubel: Made that change

PR #83, opened 2024-07-16
Retry Dimensions pubs query

lwrubel:
We have seen the `dimensions_harvest_pubs` task fail with 502 and 408 errors. Trying to automatically retry several times in those cases. 

Tested locally running task.

### DIFF
--- a/test/harvest/test_merge_pubs.py
+++ b/test/harvest/test_merge_pubs.py
@@ -138,7 +138,7 @@ def sul_pubs_csv(tmp_path):
 
 def test_dimensions_pubs_df(dimensions_pubs_csv):
     lazy_df = merge_pubs.dimensions_pubs_df(dimensions_pubs_csv)
-    assert type(lazy_df) == pl.lazyframe.frame.LazyFrame
+    assert isinstance(lazy_df, pl.lazyframe.frame.LazyFrame)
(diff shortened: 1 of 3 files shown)


test/harvest/test_merge_pubs.py
```diff
@@ -138,7 +138,7 @@ def sul_pubs_csv(tmp_path):
 
 def test_dimensions_pubs_df(dimensions_pubs_csv):
     lazy_df = merge_pubs.dimensions_pubs_df(dimensions_pubs_csv)
-    assert type(lazy_df) == pl.lazyframe.frame.LazyFrame
+    assert lazy_df.isinstance(pl.lazyframe.frame.LazyFrame)
```
lwrubel, line 141: CI runs latest version of Ruff and flagged these linting errors from new rules.

edsu (approved):

test/harvest/test_merge_pubs.py
edsu, line 141: I think this needs to be `isinstance(lazy_df, pl.lazyframe.frame.LazyFrame)`?

edsu (changes requested):

edsu (approved):

PR #88, opened 2024-07-19
Added csv_to_parquet()

edsu:
The `utils.csv_to_parquet()` function will efficiently create a Parquet file for the given CSV. It currently assumes that the Parquet file schema where all the columns are strings. But if it's useful we could adjust this and allow a schema to be passed in?

The reason why we don't simply use Polars for this is that it seems to use large amounts of memory given the CSVs that we are collecting. See https://github.com/pola-rs/polars/issues/13458#issuecomment-2238905606 for context.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -11,7 +11,8 @@ dependencies = [
     "dimcli",
     "polars>=1.2",
     "pyalex",
-    "more-itertools"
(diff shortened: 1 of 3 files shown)


edsu:
I don't think this is still relevant.

PR #89, opened 2024-07-22
Fix openalex test

lwrubel:
This test had started failing, probably because of OpenAlex author data cleanup. Updating the test to use a known Stanford author with a large number of works, and testing paging based on our hardcoded `per_page` setting of 200.


edsu (approved):

edsu:
It's nice that we don't have to worry about this one anymore!

PR #90, opened 2024-07-23
Remove authors from output Parquet

edsu:
It appears to help conserve RAM (13GB instead of 18GB) to not serialize the authorships and authors columns to the Parquet file. Since we aren't using them in any of our reporting this will help us when generating the full dataset in production.


lwrubel (approved):
Approved. @jacobthill: wanted to draw your attention to this tweak we're trying, which preserves authorship info in the publications CSVs, but would remove it from the contributions output.

lwrubel:
I think this just needs a rebase to pick up the fix to the test and the upgrade to the new version of polars.

PR #93, opened 2024-07-26
Upgrade Airflow and dependencies

lwrubel:
Seems like a good time to get Airflow and dependencies updated and see what that process might involve for the future. 

I reviewed the Airflow suggested `docker-compose.yaml` for anything new with the latest release, but nothing more was involved there than updating the Airflow version number. 

To upgrade Python dependencies:

` uv pip compile pyproject.toml -o requirements.txt --upgrade`

Tested with unit tests and also running the DAG locally.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM apache/airflow:2.9.2-python3.12
+FROM apache/airflow:2.9.3-python3.12
 
 USER root
 RUN apt-get update && apt-get install -y gcc git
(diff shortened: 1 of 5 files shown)


Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM apache/airflow:2.9.2-python3.12
+FROM apache/airflow:2.9.3-python3.12
```
edsu, line 1: I was wondering if it would be living too dangerously to install the latest version?

edsu (approved):

Dockerfile
lwrubel: Upgrading Airflow feels like upgrading Rails to me. Do we upgrade patch version automatically there but not minor?

lwrubel: I know some upgrades are a more manual process.

lwrubel: I'll go ahead and merge this but open to figuring out how to loosen up the Airflow upgrade process.

PR #98, opened 2024-08-29
Added openalex and dimensions publication links

edsu:
It might be useful to include links to the article for (possibly) getting full-text. These could be helpful when using the dataset as a source for testing machine learning techniques that use the full text as an input.

* OpenAlex: `best_oa_location` https://docs.openalex.org/api-entities/works/work-object#best_oa_location
* Dimensions: `linkout` https://docs.dimensions.ai/dsl/datasource-publications.html

### DIFF
--- a/rialto_airflow/harvest/openalex.py
+++ b/rialto_airflow/harvest/openalex.py
@@ -127,6 +127,7 @@ FIELDS = [
     "apc_paid",
     "best_oa_location",
     "biblio",
+    "citation_normalized_percentile",
     "cited_by_api_url",
     "cited_by_count",
     "cited_by_percentile_year",
--- a/test/harvest/test_merge_pubs.py
+++ b/test/harvest/test_merge_pubs.py
@@ -19,6 +19,7 @@ def dimensions_pubs_csv(tmp_path):
             "doi",
             "funders",
             "funding_section",
+            "linkout",
             "open_access",
             "publisher",
             "research_orgs",
@@ -37,6 +38,7 @@ def dimensions_pubs_csv(tmp_path):
                 "10.0000/aAaA",
                 "[]",
                 "[]",
+                "https://example.com/my-awesome-paper",
                 "True",
                 "publisher",
                 "[]",
@@ -55,6 +57,7 @@ def dimensions_pubs_csv(tmp_path):
                 "10.0000/1234",
                 "[]",
                 "[]",
+                "https://example.com/yet-another-awesome-paper",
                 "True",
                 "publisher",
                 "[]",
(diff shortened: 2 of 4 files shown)


rialto_airflow/harvest/openalex.py
```diff
@@ -127,6 +127,7 @@ def normalize_publication(pub) -> dict:
     "apc_paid",
     "best_oa_location",
     "biblio",
+    "citation_normalized_percentile",
```
edsu, line 130: This was a new column that is now coming back from OpenAlex and needed to be accounted for in the test.

test/harvest/test_merge_pubs.py
```diff
@@ -93,6 +97,7 @@ def openalex_pubs_csv(tmp_path):
                 "A Publication",
                 "article",
                 "https://doi.org/10.0000/cccc",
+                '{is_oa: true, landing_page_url: "https://www.ncbi.nlm.nih.gov/pmc/articles/PMC1398957", pdf_url: null, source: { id: "https://openalex.org/S2764455111", display_name: "PubMed Central", issn_l: null, issn: null, host_organization: "https://openalex.org/I1299303238", type: "repository" }, license: null, version: "publishedVersion"}',
```
edsu, line 100: THe value for `best_oa_location` is a JSON object.

test/harvest/test_merge_pubs.py
```diff
@@ -186,7 +193,7 @@ def test_merge(tmp_path, sul_pubs_csv, openalex_pubs_csv, dimensions_pubs_csv):
     assert output.is_file(), "output file has been created"
     df = pl.read_parquet(output)
     assert df.shape[0] == 5
-    assert df.shape[1] == 23
+    assert df.shape[1] == 25
```
edsu, line 196: OpenAlex output includes two new columns `best_oa_location` and `citation_normalized_percentile`.

test/harvest/test_merge_pubs.py
lwrubel: Will this end up as a string in our Parquet file? Maybe in a future step we could extract the fields that look useful into columns.

lwrubel (approved):

lwrubel:
Approved, pending seeing how memory usage goes when you run the DAG.

test/harvest/test_merge_pubs.py
edsu: Yes, it will be a python string representation for a dictionary unfortunately, which we will need to eval.

If we switched to harvesting as JSONL then Parquet could store it natively which would be easier to work with I think.

edsu:
> Approved, pending seeing how memory usage goes when you run the DAG.

It looks like it ran ok, but I would like to do some follow on work to unpack the JSON objects into separate columns.

PR #99, opened 2024-09-26
Add new authors.csv, adjust published columns and tests

jacobthill:
@peetucket made some changes to the RIALTO orgs app that adds `primary_school` and `primary_department` to the `authors.csv` file and changes how orgs are assigned to researchers. This PR adds the new authors.csv, publishes those columns, adds the `openalex_open_access` column, and removes some unneeded columns.


jacobthill:
I noticed `test_openalex.py::test_dois_from_orcid_paging` and `test_openalex.py::test_publications_from_dois` are failing as well but I don't think that's related to any changes in this PR.

rialto_airflow/harvest/dimensions.py
```diff
@@ -16,10 +16,10 @@
 def dois_from_orcid(orcid):
     logging.info(f"looking up dois for orcid {orcid}")
     q = """
-        search publications where researchers.orcid_id = "{}"
+        search publications where researchers.orcid_id = "{}" and year in [2018:{}]
```
edsu: Why are we limiting to publications between 2018-2024. This seems like something we will forget about, and we will quietly stop getting new data?

rialto_airflow/harvest/merge_pubs.py
```diff
@@ -86,7 +82,7 @@ def openalex_pubs_df(openalex_pubs):
             "publication_year",
             "title",
             "type",
-            "best_oa_location",
```
edsu, line 89: We found it useful to return this in the Easy Deposit workcycle, because we wanted to get the URLs for OpenAccess PDFs. But I guess it can be removed. Did you find it was not needed or redundant?

edsu (approved):

rialto_airflow/harvest/merge_pubs.py
jacobthill: My thinking was that this DAG would drive the open access dashboard and its not needed for that dashboard. I think we will want another DAG for a publishers data dashboard and others for other purposes. The other option would be to get all of the fields for all dashboards in one DAG and then ignore the fields we don't need but I recall us having concerns about the size of the file with that apporach.

rialto_airflow/harvest/dimensions.py
jacobthill: The thought was that since Rochelle only cared about pubs after 2017, we could speed up the run time by filtering in the query but I don't think it actually did speed things up so I pulled it out.

rialto_airflow/harvest/merge_pubs.py
edsu: It seems like we should minimize the time we have to go out and get metadata from OpenAlex and Dimensions. As long as we have the data in the interim files there will be an opportunity to use it other datasets like you said.

PR #154, opened 2025-02-13
tweaks to get sqlalchemy dependency and apache-airflow dev dependency working for running airflow locally

jmartin-sul:
* lined up compatible versions of sqlalchemy and apache-airflow, per https://github.com/sul-dlss/rialto-airflow/pull/153 and standup discussion yesterday
* add a package to the list of debian packages installed in our rialto-airflow docker image, per https://stackoverflow.com/a/78692169
* recompile requirements.txt with uv since pyproject.toml dep listing was updated

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -107,7 +107,7 @@ sphinxcontrib-qthelp==2.0.0
     # via sphinx
 sphinxcontrib-serializinghtml==2.0.0
     # via sphinx
-sqlalchemy==2.0.38
+sqlalchemy==1.4.54
     # via rialto-airflow (pyproject.toml)
 stack-data==0.6.3
     # via ipython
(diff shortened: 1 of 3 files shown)


jmartin-sul:
i flailed for a bit on my failing `docker compose up` call, thinking it was the usual under-resourcing of my local docker setup, since airflow is pretty resource hungry.  but when that didn't work, looking more closely into the error messages and doing some web searching turned up these fixes, which got my local instance working.

requirements.txt
```diff
@@ -107,7 +107,7 @@ sphinxcontrib-qthelp==2.0.0
     # via sphinx
 sphinxcontrib-serializinghtml==2.0.0
     # via sphinx
-sqlalchemy==2.0.38
+sqlalchemy==1.4.54
```
edsu, line 110: Good find!

edsu (approved):

requirements.txt
jmartin-sul: that was the result of trying to run things a couple times, and being confused as to why i was still getting complaints about incompatible versions, even after updating the `pyproject.toml` file 😄

Issue #132, opened 2025-02-06
Add rialto-airflow to dependency updates


peetucket:
Add to both ruby dependency update process and python dependency update process

(see speech to text repo for recent python example)

PR #156, opened 2025-02-14
add .autoupdate/preupdate hook so that our weekly dep updates can handle python deps for the project

jmartin-sul:
hooks into https://github.com/sul-dlss/access-update-scripts/blob/7c998d85ef4e27fcfee9d5601f38893deb807506/autupdate.sh#L37-L42

see e.g. https://github.com/sul-dlss/speech-to-text/pull/81/files#diff-0a2ddddfdd28a062dadabf7ea9f86285bacbb09f2d9e53da056044905e673819

connects with https://github.com/sul-dlss/rialto-airflow/issues/132

### DIFF
--- a/.autoupdate/preupdate
+++ b/.autoupdate/preupdate
@@ -1,0 +1,17 @@
+#!/bin/bash
+
+# This script is called by our weekly dependency update job in Jenkins
+
+pip3 install uv > rialto-airflow.txt &&
+    ~/.local/bin/uv lock --upgrade >> rialto-airflow.txt &&     # Currently, local dev and CI use the uv.lock file for dependency resolution...
+    uv pip compile pyproject.toml -o requirements.txt --upgrade # ...but the Dockerfile for our Airflow image uses requirements.txt.
+
+retVal=$?
+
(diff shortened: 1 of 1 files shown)


.autoupdate/preupdate
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+
+# This script is called by our weekly dependency update job in Jenkins
+
+pip3 install uv > rialto-airflow.txt &&
+    ~/.local/bin/uv lock --upgrade >> rialto-airflow.txt
```
jmartin-sul, line 6: add a third line like `uv pip compile pyproject.toml -o requirements.txt --upgrade` from readme so that we also generate an updated `requirements.txt`

.autoupdate/preupdate
```diff
@@ -0,0 +1,16 @@
...
+
+retVal=$?
+
+git add uv.lock &&
```
jmartin-sul, line 10: and then add `requirements.txt` also

edsu (approved):

PR #173, opened 2025-02-20
Adds mais-utils, starts publish-orcid DAG

jacobthill:
closes https://github.com/sul-dlss/rialto-airflow/issues/143


edsu (changes requested):
I'm honestly not a fan of `utils` modules -- if you squint right everything looks like a util ;-) So I'd like to see us not expand that namespace further if possible.

Instead of `rialto_airflow.utils.mais_utils` why don't you add it as `rialto_airflow.mais` and leave the existing utils where it is?

Also, if we can't access the MAIS API from where the Github actions run (outside of the Stanford network) we will want to mark those tests as skipped when they run in CI. Take a look at how we do this for sulpub tests https://github.com/sul-dlss/rialto-airflow/blob/main/test/harvest/test_sul_pub.py#L11-L17

rialto_airflow/dags/publish-orcid.py
edsu, line 1: I think it'll be a good idea to use an underscore instead of a hyphen when naming the module so it can be imported by Python? I'm kind of surprised Airflow is able to import it.

rialto_airflow/dags/publish_orcid.py
```diff
@@ -0,0 +1,51 @@
...
+mais_client_secret = Variable.get("mais_secret")
+
+
+@dag(
+    # schedule=@weekly,
+    start_date=datetime.datetime(2024, 1, 1),
+    catchup=False,
+)
+def publish_orcid():
+    # @task()
```
edsu, line 24: Why are parts of the code commented out? If we're not yet using it I suggest we remove it for now and add it back when it is needed.

PR #181, opened 2025-02-24
Handle author load integrity error

### LINKED ISSUE #177: Handle duplicate orcids
If duplicate ORCIDs are encountered in an authors load, log them and proceed with loading the rest of the records. The current authors.csv file has at least one instance of a duplicate ORCID. (See https://github.com/sul-dlss/rialto-orgs/issues/95)

lwrubel:
Resolves #177 to skip and log authors who have a duplicate ORCID or other duplicate data in fields with a unique constraint. 

This is constructing the session with each row of the CSV, so that each author is loaded one by one, and thus if an IntegrityError is raised, it does not affect authors earlier or later in the CSV. If there's a better way to handles sessions in this situation, would be willing to do differently. 

I had been encountering the following error when testing a situation that would raise the exception (`test_load_
dupe_orcid`). The tests were passing. 

```
Exception during reset or similar
Traceback (most recent call last):
  File "/Users/lwrubel/sul-dlss/rialto-airflow/.venv/lib/python3.12/site-packages/sqlalchemy/pool/base.py", line 763, in _finalize_fairy
    fairy._reset(pool, transaction_was_reset)
  File "/Users/lwrubel/sul-dlss/rialto-airflow/.venv/lib/python3.12/site-packages/sqlalchemy/pool/base.py", line 1038, in _reset
    pool._dialect.do_rollback(self)
  File "/Users/lwrubel/sul-dlss/rialto-airflow/.venv/lib/python3.12/site-packages/sqlalchemy/engine/default.py", line 683, in do_rollback
    dbapi_connection.rollback()
psycopg2.OperationalError: server closed the connection unexpectedly
	This probably means the server terminated abnormally
	before or while processing the request.
```

I added [pessimistic testing of connections](https://docs.sqlalchemy.org/en/14/core/pooling.html#disconnect-handling-pessimistic) in test (we can see if we need this in production but I haven't noticed it being needed so far when looking at logs in the DAG).  I also needed to add an explicit close of sessions in `conftest.py` (`session.close_all_sessions()`) based on the discussion at the end of[ this sqlalchemy issue](https://github.com/sqlalchemy/sqlalchemy/issues/5522). The combination of these two changes eliminated the postgres OperationalErrors.

### DIFF
--- a/rialto_airflow/database.py
+++ b/rialto_airflow/database.py
@@ -18,18 +18,18 @@ def db_uri(database_name):
     return f"{os.environ.get('AIRFLOW_VAR_RIALTO_POSTGRES')}/{database_name}"
 
 
-def engine_setup(database_name: str):
+def engine_setup(database_name: str, pool_pre_ping=False):
     """
     When creating the database and its schema, use an engine and connection.
     Subsequent querying should be done by accessing an engine via get_engine.
     """
-    return create_engine(db_uri(database_name), echo=True)
+    return create_engine(db_uri(database_name), pool_pre_ping=pool_pre_ping, echo=True)
 
 
 @cache
 def get_engine(database_name: str):
     """Memoized engine for use in other modules"""
-    return engine_setup(database_name)
+    return engine_setup(database_name, pool_pre_ping=True)
 
 
 def create_database(database_name: str) -> str:
--- a/rialto_airflow/harvest/authors.py
+++ b/rialto_airflow/harvest/authors.py
@@ -8,7 +8,7 @@ from rialto_airflow.database import get_engine, Author
 from rialto_airflow.utils import rialto_authors_file
 
 
-def load_authors_table(snapshot) -> str:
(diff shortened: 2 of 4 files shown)


rialto_airflow/database.py
```diff
@@ -18,18 +18,18 @@ def db_uri(database_name):
...
 
 
-def engine_setup(database_name: str):
+def engine_setup(database_name: str, pool_pre_ping=False):
     """
     When creating the database and its schema, use an engine and connection.
     Subsequent querying should be done by accessing an engine via get_engine.
     """
-    return create_engine(db_uri(database_name), echo=True)
+    return create_engine(db_uri(database_name), pool_pre_ping=pool_pre_ping, echo=True)
```
lwrubel, line 26: Pessimistic testing of connections. Only set to True when creating an engine in testing.

edsu (approved):
I'm glad you were able to track down the intermittent testing problem.

PR #183, opened 2025-02-24
create normalize_orcid function

### LINKED ISSUE #151: Create and use a normalize_orcid function
We are normalizing ORCIDs in a few different places and it would be helpful to have one function in `rialto_airflow.utils` that did it:

https://github.com/sul-dlss/rialto-airflow/blob/6bb011029ac50f699f29a89ed9d92e033b7cfc2a/rialto_airflow/harvest/openalex.py#L65

https://github.com/sul-dlss/rialto-airflow/blob/6bb011029ac50f699f29a89ed9d92e033b7cfc2a/rialto_airflow/harvest/dimensions.py#L44

https://github.com/sul-dlss/rialto-airflow/blob/6bb011029ac50f699f29a89ed9d92e033b7cfc2a/rialto_airflow/harvest/openalex.py#L26

peetucket:
Fixes #151 - create and use a normalize_orcid function

Also, update README to reflect updated environment name and use the existing normalize_doi function in a spot where it wasn't being used

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,14 +60,14 @@ done
 ```
 
 5. The harvest DAG requires a CSV file of authors from rialto-orgs to be available. This is not yet automatically available, so to set up locally, download the file at
-https://sul-rialto-stage.stanford.edu/authors?action=index&commit=Search&controller=authors&format=csv&orcid_filter=&q=. Put the `authors.csv` file in the `data/` directory. 
+https://sul-rialto-stage.stanford.edu/authors?action=index&commit=Search&controller=authors&format=csv&orcid_filter=&q=. Put the `authors.csv` file in the `data/` directory.
 
-6. Bring up containers. 
+6. Bring up containers.
 ```
 docker compose up -d
 ```
 
-7. The Airflow application will be available at `localhost:8080` and can be accessed with the default Airflow username and password. 
+7. The Airflow application will be available at `localhost:8080` and can be accessed with the default Airflow username and password.
 
 ## Development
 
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -107,8 +107,10 @@ DOCKER_DEFAULT_PLATFORM="linux/amd64" docker build . -t suldlss/rialto-airflow:l
...
 ```
 
-Deployment to https://sul-rialto-airflow-dev.stanford.edu/ is handled like other SDR services using Capistrano. You'll need to have Ruby installed and then:
+Deployment to https://sul-rialto-airflow-XXXX.stanford.edu/ is handled like other SDR services using Capistrano. You'll need to have Ruby installed and then:
 
 ```
-bundle exec cap dev deploy
+bundle exec cap stage deploy # stage
+bundle exec cap prod deploy  # prod
+# Note: there is no QA
```
peetucket, line 115: fix environment names, formatting updates

rialto_airflow/utils.py
```diff
@@ -49,3 +49,12 @@ def normalize_doi(doi):
     doi = re.sub("^doi: ", "", doi)
 
     return doi
+
+
+def normalize_orcid(orcid):
+    orcid = orcid.strip().lower()
+    orcid = orcid.replace("https://orcid.org/", "").replace(
+        "https://sandbox.orcid.org/", ""
+    )
```
peetucket, line 58: sandbox.orcid.org is their test environment -- it is possible we will run into ORCIDs like this if using the ORCID test system (e.g. https://sandbox.orcid.org/0000-0002-7262-6251)

Issue #158, opened 2025-02-14
add code coverage (codecov) and type checking (mypy)


jmartin-sul:
- [x] review/fix/merge #185
- [x] add coverage as a required PR check
  - [x] https://github.com/sul-dlss -- `Settings` -> `Third Party Access` : `Github Apps` -> `Codecov` : `Configure` -> `Repository Access` -> add this repo to the selected repositories list
  - [x] push a new commit to an existing PR, or create a new PR, so that Codecov reports to said PR after CI completes and reports coverage results to codecov
  - [x] https://github.com/sul-dlss/rialto-airflow -- `Settings` -> `Code and automation` : `Branches` -> `main` : `Edit` -> ✅   `Require status checks to pass before merging` -> use search box to add Codecov to `Status checks that are required`

jmartin-sul:
done once #185 is merged, and coverage can be added as a required PR check

jmartin-sul:
done
<img width="972" alt="Image" src="https://github.com/user-attachments/assets/55c8b249-00bd-4aaa-bc6f-3010c572495f" />

PR #185, opened 2025-02-25
add code coverage and type checking and a few other touchups

jmartin-sul:
* where possible/required, fix issues discovered by type checking, including updating some function calls to match current signatures.  where necessary, ignore spurious errors flagged by type checker confusion.
* a few bits of explicit type specification added opportunistically near places where the type checker actually called for it, if it felt like it added readability
* a bit of variable renaming for clarity
* use dict and list builtins for type info, instead of deprecated proxies from typing package
* update readme with info about code coverage and type checking
* add coverage reporting and type checking to CI
* move CI linter to official ruff github action instead of deprecated 3rd party version

connects with #158

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -18,14 +18,11 @@ jobs:
           python-version: ${{ matrix.python-version }}
 
       - name: Lint
-        uses: chartboost/ruff-action@v1
-        # Default action is 'check'
-        # it may move, see https://github.com/astral-sh/ruff/issues/8400
-
-      - name: Format
-        uses: chartboost/ruff-action@v1
+        uses: astral-sh/ruff-action@v3
         with:
-          args: 'format --check'
+          # if files would be reformatted: print the diff between the current file and
+          # the re-formatted file, then exit with non-zero status code (which should fail the build step)
+          args: 'format --diff'
 
       - name: Install uv
         run: pip3 install uv
@@ -33,8 +30,17 @@ jobs:
       - name: Bring up postgres container
         run: docker compose -f compose.yaml up -d --wait postgres
 
+      - name: Run type checking
+        run: uv run mypy .
+
       - name: Run tests
-        run: uv run pytest
+        run: uv run pytest --cov-branch --cov-report=xml
         env:
           AIRFLOW_VAR_DIMENSIONS_API_USER: ${{ secrets.AIRFLOW_VAR_DIMENSIONS_API_USER }}
           AIRFLOW_VAR_DIMENSIONS_API_PASS: ${{ secrets.AIRFLOW_VAR_DIMENSIONS_API_PASS }}
+
+      - name: Upload coverage reports to Codecov
+        uses: codecov/codecov-action@v5
+        with:
+          token: ${{ secrets.CODECOV_TOKEN }}
+          slug: sul-dlss/rialto-airflow
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -22,20 +22,34 @@ dependencies = [
     "pyalex",
     "more-itertools",
     "psycopg2-binary>=2.9.10",
-    "sqlalchemy>=1.4.36,<2.0",
+    "sqlalchemy>=1.4.36,<2.0", # airflow still incompatible with 2.x, see https://github.com/apache/airflow/issues/28723
 ]
 
 [tool.pytest.ini_options]
 pythonpath = ["."]
-addopts = "-v"
 markers = "mais_tests: Tests requiring MAIS access"
+addopts = "-v --cov --cov-report=html --cov-report=term"
+
+[tool.mypy]
+check_untyped_defs = true # Type-checks the interior of functions without type annotations.
+
+[[tool.mypy.overrides]]
+module = ["dimcli", "pyalex", "requests_oauthlib", "sqlalchemy_utils"]
+ignore_missing_imports = true
 
 [dependency-groups]
 dev = [
     "apache-airflow==2.10.4", # aligned with base Docker image
+    "mypy>=1.15.0",
     "pytest>=8.3.4",
+    "pytest-cov>=6.0.0",
     "python-dotenv>=1.0.1",
     "ruff>=0.9.4",
+
+    # packages with type info, for mypy
+    "pandas-stubs",
+    "sqlalchemy-stubs>=0.4", # when airflow allows SQLAlchemy >= 2.x, see if this is still needed/compatible
+    "types-requests",
 ]
 
 [build-system]
--- a/rialto_airflow/dags/harvest.py
+++ b/rialto_airflow/dags/harvest.py
@@ -21,10 +21,13 @@ sul_pub_host = Variable.get("sul_pub_host")
 sul_pub_key = Variable.get("sul_pub_key")
 
 # to artificially limit the API activity in development
+dev_limit = None
 try:
     dev_limit = int(Variable.get("dev_limit", default_var=None))
 except TypeError:
-    dev_limit = None
+    pass
+except ValueError:
+    pass
 
 
 @dag(
@@ -60,8 +63,9 @@ def harvest():
         """
         Fetch the data by ORCID from Dimensions.
         """
+        authors_csv = snapshot.path / "authors.csv"
         pickle_file = snapshot.path / "dimensions-doi-orcid.pickle"
-        dimensions.doi_orcids_pickle(pickle_file, limit=dev_limit)
+        dimensions.doi_orcids_pickle(authors_csv, pickle_file, limit=dev_limit)
         return str(pickle_file)
 
     @task()
@@ -69,8 +73,9 @@ def harvest():
         """
         Fetch the data by ORCID from OpenAlex.
         """
+        authors_csv = snapshot.path / "authors.csv"
         pickle_file = snapshot.path / "openalex-doi-orcid.pickle"
-        openalex.doi_orcids_pickle(pickle_file, limit=dev_limit)
+        openalex.doi_orcids_pickle(authors_csv, pickle_file, limit=dev_limit)
         return str(pickle_file)
 
     @task()
@@ -89,8 +94,9 @@ def harvest():
         Extract a mapping of DOI -> [SUNET] from the dimensions doi-orcid dict,
         openalex doi-orcid dict, SUL-Pub publications, and authors data.
         """
+        authors_csv = snapshot.path / "authors.csv"
         pickle_file = snapshot.path / "doi-sunet.pickle"
-        create_doi_sunet_pickle(dimensions, openalex, sul_pub, authors, pickle_file)
+        create_doi_sunet_pickle(dimensions, openalex, sul_pub, authors_csv, pickle_file)
 
         return str(pickle_file)
 
@@ -128,10 +134,11 @@ def harvest():
         """
         Get contributions from publications.
         """
-        output = snapshot.path / "contributions.parquet"
-        create_contribs(pubs, doi_sunet_pickle, output)
+        authors_csv = snapshot.path / "authors.csv"
+        contribs_path = snapshot.path / "contributions.parquet"
+        create_contribs(pubs, doi_sunet_pickle, authors_csv, contribs_path)
 
-        return str(output)
+        return str(contribs_path)
 
     @task()
     def publish(pubs_to_contribs, merge_publications):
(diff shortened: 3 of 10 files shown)


.github/workflows/test.yml
```diff
@@ -18,23 +18,29 @@ jobs:
...
-        # it may move, see https://github.com/astral-sh/ruff/issues/8400
-
-      - name: Format
-        uses: chartboost/ruff-action@v1
+        uses: astral-sh/ruff-action@v3
         with:
-          args: 'format --check'
+          # if files would be reformatted: print the diff between the current file and
+          # the re-formatted file, then exit with non-zero status code (which should fail the build step)
+          args: 'format --diff'
```
jmartin-sul, line 25: switch from deprecated to officially supported action

rialto_airflow/dags/harvest.py
```diff
@@ -21,10 +21,11 @@
 sul_pub_key = Variable.get("sul_pub_key")
 
 # to artificially limit the API activity in development
+dev_limit = None
 try:
     dev_limit = int(Variable.get("dev_limit", default_var=None))
 except TypeError:
-    dev_limit = None
+    pass
```
jmartin-sul, line 28: one of the few hoops i jumped through to make things less confusing for the type checker (otherwise it thinks `dev_limit` should be a non-optional `int`, since that was its first possible assignment).

rialto_airflow/dags/harvest.py
```diff
@@ -60,17 +61,19 @@ def dimensions_harvest_dois(snapshot):
         """
         Fetch the data by ORCID from Dimensions.
         """
+        authors_csv = snapshot.path / "authors.csv"
         pickle_file = snapshot.path / "dimensions-doi-orcid.pickle"
-        dimensions.doi_orcids_pickle(pickle_file, limit=dev_limit)
+        dimensions.doi_orcids_pickle(authors_csv, pickle_file, limit=dev_limit)
```
jmartin-sul, line 68: these `authors_csv` changes are one of the parts of this i'm least sure about, but afaict, this is a legit signature mismatch that the type checker caught, and it doesn't seem to be covered by tests?  it seemed to me that the path was what was wanted and not the csv content, but i walso wasn't 100% sure about that.

rialto_airflow/dags/harvest.py
```diff
@@ -128,10 +132,11 @@ def pubs_to_contribs(pubs, doi_sunet_pickle, snapshot):
         """
         Get contributions from publications.
         """
-        output = snapshot.path / "contributions.parquet"
-        create_contribs(pubs, doi_sunet_pickle, output)
+        authors_csv = snapshot.path / "authors.csv"
+        contribs_path = snapshot.path / "contributions.parquet"
```
jmartin-sul, line 138: this seemed like a clearer name, and lined up better with the param name of the `create_contribs` function

rialto_airflow/database.py
```diff
@@ -6,7 +6,7 @@
 from sqlalchemy import create_engine, text
 from sqlalchemy.dialects.postgresql import ARRAY, JSONB
 from sqlalchemy.ext.compiler import compiles
-from sqlalchemy.orm import declarative_base, relationship
+from sqlalchemy.orm import RelationshipProperty, declarative_base, relationship  # type: ignore
```
jmartin-sul, line 9: `declarative_base` is the thing being ignored here

rialto_airflow/database.py
```diff
@@ -69,7 +69,7 @@ def pg_utcnow(element, compiler, **kw):
 )
 
 
-class Publication(Base):
+class Publication(Base):  # type: ignore
```
jmartin-sul, line 72: there's some magic used in `Base` that trips up mypy.  if/when airflow becomes sqlalchemy 2.x compatible, this should no loner be needed, because sqlalchemy 2.x is supposed to play better with type checkers.  (i think sqlalchemy 1.x predates widespread typechcker usage?)

rialto_airflow/database.py
```diff
@@ -83,12 +83,12 @@ class Publication(Base):
     pubmed_json = Column(JSONB)
     created_at = Column(DateTime, server_default=utcnow())
     updated_at = Column(DateTime, onupdate=utcnow())
-    authors = relationship(
+    authors: RelationshipProperty = relationship(
```
jmartin-sul, line 86: these join tables seem to be something mypy has trouble inferring type on.  now wondering whether it's possible to give this more precision `Author` as the type on the other side.  i only added this because the type checker pointed to this as one of the couple spots that required type clarification.

rialto_airflow/database.py
```diff
@@ -107,7 +107,7 @@ class Author(Base):
     primary_division = Column(String)
     created_at = Column(DateTime, server_default=utcnow())
     updated_at = Column(DateTime, onupdate=utcnow())
-    publications = relationship(
+    publications: RelationshipProperty = relationship(
```
jmartin-sul, line 110: the other spot that require clarification from the type checker

rialto_airflow/harvest/authors.py
```diff
@@ -20,7 +20,7 @@ def load_authors_table(snapshot) -> str:
     logging.info(
         f"Loading authors from {authors_file} into database {snapshot.database_name}"
     )
-    with Session.begin() as session:
+    with Session.begin() as session:  # type: ignore
```
jmartin-sul, line 23: instantiation of some sqlalchemy building block confuses the type checker again (a running theme 😆)

rialto_airflow/harvest/authors.py
```diff
@@ -75,7 +75,7 @@ def to_boolean(value: str) -> bool:
     bool_map = {"true": True, "false": False}
 
     # will raise KeyError if unexpected value
-    return bool_map.get(value.strip().lower())
+    return bool_map[value.strip().lower()]
```
jmartin-sul, line 78: turns out `get` can return `None`, and the square brackets will give the `KeyError` if the thing is missing (also tested in the REPL)

rialto_airflow/harvest/doi_sunet.py
```diff
@@ -10,7 +10,7 @@
 
 def create_doi_sunet_pickle(
     dimensions: str, openalex: str, sul_pub_csv: str, authors_csv: str, output_path
-) -> dict:
+) -> None:
```
jmartin-sul, line 13: doesn't actually seem to return anything

rialto_airflow/harvest/doi_sunet.py
```diff
@@ -89,21 +89,23 @@ def get_author_maps(authors):
     df["orcidid"] = df["orcidid"].apply(orcid_id)
 
     # orcid -> sunet
-    orcid = pd.Series(df["sunetid"].values, index=df["orcidid"]).to_dict()
+    orcid: dict[str, str] = pd.Series(
```
jmartin-sul, line 92: the `pd.Series()` results were the parts where the type checker asked for a type to be specified.  the rest of the added type specifications in this file just seemed to make things clearer at a glance to me, and passed the type checker.

rialto_airflow/mais.py
```diff
@@ -2,7 +2,7 @@
 import logging
 from collections import defaultdict
 from datetime import date
-from typing import Any, Dict, List, Optional, Union
+from typing import Any, Optional, Union
```
jmartin-sul, line 5: turns out `typing.Dict` and `typing.List` are deprecated in favor of plain `dict` and `list` (nothing complained about that, i stumbled on it in the docs)

rialto_airflow/mais.py
```diff
@@ -31,7 +31,11 @@
 logger = logging.getLogger(__name__)
 
 
-def get_token(client_id: str, client_secret: str, base_url: str) -> Optional[str]:
+class TokenFetchError(LookupError):
+    pass
```
jmartin-sul, line 35: waffled between adding this and just using `LookupError` directly.  happy to ditch the custom error if that's more pythonic or otherwise preferable.

rialto_airflow/mais.py
```diff
@@ -58,21 +62,20 @@ def get_token(client_id: str, client_secret: str, base_url: str) -> Optional[str
             logger.info("Successfully obtained access token.")  # Log success
             return access_token
         else:
-            logger.error(
-                f"No 'access_token' found in JSON response: {data}"
-            )  # Log specific error
-            return None
+            err_msg = f"No 'access_token' found in JSON response: {data}"
+            logger.error(err_msg)  # Log specific error
+            raise TokenFetchError(err_msg)
```
jmartin-sul, line 67: this sprang from the type checker complaining that a subsequent call[^1] treated the return value from this function as if it would be non-`None`.  could've made the value optional and raised in the receiving function, but a cursory read through of the code made me think that a bunch of downstream stuff assumed success here, so i figured maybe better to just fail early?  happy to revisit if `return None` was better for reasons i didn't understand.

[^1]:  `current_orcid_users` in `orcid_integration_stats` task in `publish_orcid` DAG/module

rialto_airflow/dags/harvest.py
```diff
@@ -21,10 +21,13 @@
...
 
 # to artificially limit the API activity in development
+dev_limit = None
 try:
     dev_limit = int(Variable.get("dev_limit", default_var=None))
 except TypeError:
-    dev_limit = None
+    pass
+except ValueError:
+    pass
```
jmartin-sul, line 30: fwiw, i got a `ValueError` (tried to call `int` on `''` before i set the variable, not sure if `None` was turned to that or what) when starting airflow locally to confirm it still worked.  since the spirit of this seemed to be resilience against an undeclared variable, this seemed like the right thing to do?

test/test_mais.py
```diff
@@ -46,14 +46,17 @@ def test_get_token_success(base_url, client_id, client_secret):
...
         pytest.skip("No MAIS credentials available")
-    token = mais.get_token("dummy_invalid_id", "dummy_invalid_secret", base_url)
-    assert token is None
+    with pytest.raises(mais.TokenFetchError):
+        mais.get_token("dummy_invalid_id", "dummy_invalid_secret", base_url)
 
 
 @pytest.mark.mais_tests
-def test_get_response_success(access_token, base_url):
+def test_get_response_success(access_token, base_url, client_id, client_secret):
```
jmartin-sul, line 59: type checker complained about mismatch, and it seemed upon comparison to other nearby tests that this (and the couple similar changes below) might've been an oversight in providing fixtures via params?

jmartin-sul (commented):
explanatory comments left inline.

didn't run any of the DAGs, but did confirm that airflow parsed them without error after rebuilding and starting my docker setup locally.

rialto_airflow/dags/harvest.py
edsu: I think the risk here is minimal since I don't think we are not going to be keeping this code.

rialto_airflow/database.py
edsu: That sounds plausible to me.

rialto_airflow/harvest/authors.py
edsu: Yes, maybe this will get easier once we can use sqlalchemy 2? 

https://docs.sqlalchemy.org/en/20/orm/extensions/mypy.html

rialto_airflow/harvest/doi_sunet.py
edsu: Yes, I think these pickles are going to go in the dustbin, so it doesn't matter too much I don't think.

rialto_airflow/mais.py
edsu: Good to know, I've seen both styles and was wondering which was preferred.

pyproject.toml
```diff
@@ -22,20 +22,34 @@ dependencies = [
...
     "apache-airflow==2.10.4", # aligned with base Docker image
+    "mypy>=1.15.0",
     "pytest>=8.3.4",
+    "pytest-cov>=6.0.0",
     "python-dotenv>=1.0.1",
     "ruff>=0.9.4",
+
+    # packages with type info, for mypy
+    "pandas-stubs",
+    "sqlalchemy-stubs>=0.4", # when airflow allows SQLAlchemy >= 2.x, see if this is still needed/compatible
```
edsu, line 51: Yeah, I saw in https://docs.sqlalchemy.org/en/20/orm/extensions/mypy.html that 2.x has been designed to play nicer with type checking?

edsu (approved):

PR #190, opened 2025-02-26
Harvest from OpenAlex to the database

### LINKED ISSUE #163: Write harvest_openalex task
The `harvest_openalex` task will replace the `openalex_harvest_dois` task currently in place. It will need to:

- Query the OpenAlex API by ORCID iD

We will no longer need the doi-orcid pickle file currently returned by `openalex_harvest_dois`.

edsu:
Closes #163


rialto_airflow/harvest/openalex.py
```diff
@@ -1,178 +1,101 @@
...
-            if orcid:
-                count += 1
-                orcid_dois[orcid] = list(dois_from_orcid(orcid))
-                if limit is not None and count > limit:
+    jsonl_file = snapshot.path / "openalex.jsonl"
+    count = 0
+    stop = False
+
+    with jsonl_file.open("w") as jsonl_output:
+        with get_session(snapshot.database_name).begin() as select_session:
```
edsu, line 35: Is it right to create two separate sqlalchemy sessions if I want to do inserts to publications while selecting from the authors table?

compose.yaml
```diff
@@ -117,6 +117,7 @@ services:
       retries: 5
       start_period: 5s
     restart: always
+    command: postgres -c 'max_connections=250'
```
edsu, line 120: up from 100

rialto_airflow/dags/harvest.py
edsu, line 1: Note: the state of the DAG in terms of task return values that are passed around don't make much sense at the moment, but hopefully as we replace the harvesting code, and remove things that are no longer needed it will start to look better.

rialto_airflow/utils.py
```diff
@@ -44,6 +44,9 @@ def invert_dict(dict):
```
edsu, line 45: Not all publications have a DOI, but we still want to store them in the Publications table. It seemed easier to just let this function return None when passed a None?

test/harvest/test_sul_pub.py
```diff
@@ -17,55 +17,6 @@
 no_auth = not (sul_pub_host and sul_pub_key)
```
edsu, line 19: Moving these fixtures to conftest.py so they can be used in other tests.

test/conftest.py
```diff
@@ -48,3 +55,62 @@ def test_session(test_engine):
         yield sessionmaker(test_engine)
     finally:
         close_all_sessions()
```
edsu, line 57: The thinking was that these fixtures could be useful for other harvesting tests.

jmartin-sul (approved):
some small questions/suggestions.  if you did end up wanting to make any changes based on the review, no preference as to whether it's in this PR or in follow up work, i'm fine with merging this as-is.

rialto_airflow/harvest/openalex.py
edsu: Thanks for pointing out our database connections are currently pulling all results back into the client. In this case it's a small number of rows (7000?) so it's not a big deal. But this will be important when we are doing the "filling in" work where we iterate through all the publications and add look for missing metadata.

Unfortunately once the Session goes out of scope the ORM Author objects no longer can be used because they are in a "detached" state. Instead I could query the `cap_profile_id` and `author_id` as a list of dictionaries. But it's nice having the Author object I think, so I'm going to leave it the way it is, at least for now.

rialto_airflow/harvest/openalex.py
```diff
@@ -1,178 +1,101 @@
...
+            # get all authors that have an ORCID
+            for author in (
+                select_session.query(Author).where(Author.orcid.is_not(None)).all()  # type: ignore
+            ):
+                if stop is True:
+                    logging.info(f"Reached limit of {limit} authors stopping")
```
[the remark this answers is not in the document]
edsu: Yes, that's right -- thanks for spotting this, it could have led to a lot of debugging confusion later.

test/conftest.py
```diff
@@ -48,3 +55,62 @@ def test_session(test_engine):
...
+        session.add(pub)
+        return pub
+
+
+@pytest.fixture
+def mock_association(test_session, mock_publication, mock_authors):
+    with test_session.begin() as session:
+        session.execute(
+            insert(pub_author_association).values(
+                # TODO: should the IDs be looked up in case they aren't always 1?
```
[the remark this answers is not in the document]
edsu: Yes, I think that's right.

edsu:
I'm going to see if I can resolve the coverage drop which is preventing the merge.

PR #191, opened 2025-02-26
Tiny authors csv refactor plus coverage tweak

jmartin-sul:
* centralize authors csv path construction into field on `Snapshot` class
  * i felt like this made things a bit more readable and centralized some path construction that we'd probably want to change across all current uses if it did change, but this was like 10 minutes of work, so i'm happy to just drop this commit if others don't think it's an improvement.
* test config touchup: test code itself shouldn't be counted in coverage stats
  * this unfortunately drops the overall coverage for the codebase, leading one codecov check to fail on this PR.  making the stats more honest seemed like an improvement by itself, even if we do want to consider upping coverage in light of the new stats, so i wasn't inclined to find some uncovered spot to test solely for the sake of this PR's checks.
* add myself to the contributors list 😄

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -30,6 +30,9 @@ pythonpath = ["."]
 markers = "mais_tests: Tests requiring MAIS access"
 addopts = "-v --cov --cov-report=html --cov-report=term"
 
+[tool.coverage.run]
+omit = ["test/*"]
+
 [tool.mypy]
 check_untyped_defs = true # Type-checks the interior of functions without type annotations.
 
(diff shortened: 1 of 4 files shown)


lwrubel:
I'm noticing that we will be putting other filenames in other tasks (e.g. the openalex harvest will be writing to an openalex.jsonl). So I'm not sure what pattern we'll want to follow for these. Open to whatever approach. We may find that we need to reference these files from other tasks or DAGs and that would also be an argument for putting them in the Snapshot class eventually.

PR #203, opened 2025-03-03
Keep unapproved sul_pub pubs for the jsonl file

### LINKED ISSUE #195: Include all publications from sul-pub in jsonl export
We want to be able to report on how people are using sulpub. To do that we will need to collect all publications from sul-pub for all status (new/denied/approved), not just the ones that have been approved by a researcher.

We are currently harvesting all the publications but only storing approved ones in the database. This makes subsequent reporting easier. If we update the sulpub harvesting task to write all publications to the JSON-L file (irrespective of whether they are approved) then we will be able to generate reports of how people are using sulpub. The advantage of not storing them all in the database is that they won't require us to remember to ignore them, since many are not by Stanford researchers.

lwrubel:
Resolves #195 to write all sul_pub publications to the jsonl files, even if not approved by any author. Only approved pubs are added to the database.

### DIFF
--- a/rialto_airflow/harvest/sul_pub.py
+++ b/rialto_airflow/harvest/sul_pub.py
@@ -25,38 +25,40 @@ def harvest(snapshot, host, key, per_page=1000, limit=None):
             with get_session(snapshot.database_name).begin() as session:
                 doi = extract_doi(sulpub_pub)
 
-                # if when inserting the row we get a constraint violation
-                # on the DOI then we have to do an update
-                pub_id = session.execute(
-                    insert(Publication)
-                    .values(doi=doi, sulpub_json=sulpub_pub)
-                    .on_conflict_do_update(
-                        constraint="publication_doi_key",
-                        set_=dict(sulpub_json=sulpub_pub),
-                    )
-                    .returning(Publication.id)
-                ).scalar_one()
+                # only write approved publications to the database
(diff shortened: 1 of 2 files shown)


rialto_airflow/harvest/sul_pub.py
```diff
@@ -25,38 +25,40 @@ def harvest(snapshot, host, key, per_page=1000, limit=None):
...
-                # if when inserting the row we get a constraint violation
-                # on the DOI then we have to do an update
-                pub_id = session.execute(
-                    insert(Publication)
-                    .values(doi=doi, sulpub_json=sulpub_pub)
-                    .on_conflict_do_update(
-                        constraint="publication_doi_key",
-                        set_=dict(sulpub_json=sulpub_pub),
+                # only write approved publications to the database
+                if approved(sulpub_pub):
```
lwrubel, line 29: Code below is not new, just moved within a conditional.

rialto_airflow/harvest/sul_pub.py
```diff
@@ -93,9 +95,6 @@ def publications(host, key, per_page=1000, limit=None):
             more = False
 
         for record in records:
-            if not approved(record):
-                continue
-
```
lwrubel, line 98: Now we keep these pubs.

edsu (approved):
Nice!

PR #204, opened 2025-03-03
Adding Web of Science harvest

### LINKED ISSUE #165: Write harvest_wos task
The `harvest_wos` will need to:

- Query the Web of Science API by ORCID iD

edsu:
This adds a harvester for Web of Science publications by ORCID. I've tested with `AIRFLOW_VAR_DEV_LIMIT=50000` which helped discover some unexpected responses from the WoS API that I needed to handle see #207 and #208.

The API call was determined by looking at [sul_pub's querying by ORCID](https://github.com/sul-dlss/sul_pub/blob/d4f18197aee461a2a67400dfc3d61fd4c1b4f93e/lib/web_of_science/query_orcid.rb#L37) and how the [API KEY is passed](https://github.com/sul-dlss/sul_pub/blob/d4f18197aee461a2a67400dfc3d61fd4c1b4f93e/lib/clarivate/rest_client.rb#L28-L32). 

Closes #165


pyproject.toml
```diff
@@ -28,7 +28,8 @@ dependencies = [
 [tool.pytest.ini_options]
 pythonpath = ["."]
 markers = "mais_tests: Tests requiring MAIS access"
-addopts = "-v --cov --cov-report=html --cov-report=term"
+addopts = "-v --cov --cov-report=html --cov-report=term --log-level INFO --log-file test.log"
```
edsu, line 31: It was useful for me to see the log messages. But I can remove if this is noisy.

rialto_airflow/dags/harvest.py
```diff
@@ -150,6 +159,9 @@ def publish(pubs_to_contribs, merge_publications):
 
     openalex_jsonl = openalex_harvest(snapshot)
```
edsu, line 161: Eventually we'll want to tie the WoS harvesting into the workflow, but for now nothing is dependent on it.

test/harvest/test_wos.py
```diff
@@ -0,0 +1,214 @@
...
+    monkeypatch.setattr(wos, "orcid_publications", f)
+
+
+@pytest.fixture
+def existing_publication(test_session):
+    with test_session.begin() as session:
+        pub = Publication(
+            doi="10.1515/9781503624153",
+            sulpub_json={"sulpub": "data"},
+            wos_json={"wos": "data"},
```
lwrubel, line 71: Is there a reason to have `wos_json` already populated in the existing_publication fixture? It seems like the assertion on L149 would always pass because of the fixture.

rialto_airflow/dags/harvest.py
```diff
@@ -74,6 +74,15 @@ def openalex_harvest(snapshot):
 
         return jsonl_file
 
+    @task()
+    def wos_harvest(snapshot):
+        """
+        Fetch the data by ORCID from OpenAlex.
```
lwrubel, line 80: Fetch from Web of Science?

lwrubel (approved):
Approved pending minor comments

test/harvest/test_wos.py
edsu: This is a good point. I'll unset it.

PR #212, opened 2025-03-05
Updated diagram

edsu:
This updates our diagram to better reflect the new (anticipated) workflow.

It's easiest to see what this looks like here: https://github.com/sul-dlss/rialto-airflow/tree/diagram#readme


jacobthill (changes requested):
Maybe the "ORCID-Adoption" box should not be included in the publish step since it uses a different DAG and a mostly different set of data.

edsu:
Maybe it makes sense to remove the diagram and rely on what Airflow gives us?

PR #217, opened 2025-03-06
Fill in OpenAlex pubs

### LINKED ISSUE #194: Create "fill-in" task for OpenAlex
After all the initial harvesting of publications by ORCID is complete we want to examine all publications that have a DOI and also lack OpenAlex metadata, to see if OpenAlex metadata is actually available when querying by DOI. The rationale for this previously was that some publications may be in OpenAlex, but may not be associated with a ORCID.

We had some code for this, which took advantage of the fact that you can query for more than one DOI at a time:

https://github.com/sul-dlss/rialto-airflow/blob/94c1a28339a5255c14691312a0dd795b7b5909b3/rialto_airflow/harvest/openalex.py#L83-L109

We will want to use the database in such a way that we aren't pulling all the query results back into memory for `SELECT * FROM publications WHERE sulpub_json IS NULL AND doi IS NOT NULL`.

It might be interesting to keep a count of how often we find metadata this way, and log it at the end, to help evaluate how important this is.

lwrubel:
Resolves #194 to fill in OpenAlex json for DOIs that lack it after the by-ORCID harvest. 

A run with a dev limit of 10000 showed that publications from openalex were filled in.

### DIFF
--- a/rialto_airflow/harvest/openalex.py
+++ b/rialto_airflow/harvest/openalex.py
@@ -117,6 +117,7 @@ def fill_in(snapshot: Snapshot, jsonl_file: Path) -> Path:
                 .where(Publication.openalex_json.is_(None))
                 .execution_options(yield_per=100)
             )
+            # TODO: consider getting data for more than one DOI at a time
             for row in select_session.execute(stmt):
                 logging.info(f"filling in data for {row.doi}")
                 try:
--- a/rialto_airflow/harvest/openalex.py
+++ b/rialto_airflow/harvest/openalex.py
@@ -111,10 +111,14 @@ def fill_in(snapshot: Snapshot, jsonl_file: Path) -> Path:
     count = 0
     with jsonl_file.open("a") as jsonl_output:
         with get_session(snapshot.database_name).begin() as select_session:
-            stmt = select(Publication.doi).where(Publication.doi.is_not(None)).where(Publication.openalex_json.is_(None)).execution_options(yield_per=100)
+            stmt = (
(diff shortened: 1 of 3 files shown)


rialto_airflow/harvest/openalex.py
```diff
@@ -102,3 +104,40 @@ def orcid_publications(orcid: str) -> Generator[dict, None, None]:
...
+        with get_session(snapshot.database_name).begin() as select_session:
+            stmt = (
+                select(Publication.doi)  # type: ignore
+                .where(Publication.doi.is_not(None))  # type: ignore
+                .where(Publication.openalex_json.is_(None))
+                .execution_options(yield_per=100)
+            )
+            for row in select_session.execute(stmt):
+                logging.info(f"filling in data for {row.doi}")
+                try:
```
edsu, line 122: If we wanted to speed up the lookups I think we could get more than one DOI at a time?

https://github.com/sul-dlss/rialto-airflow/blob/94c1a28339a5255c14691312a0dd795b7b5909b3/rialto_airflow/harvest/openalex.py#L87

edsu (approved):
I left a comment about batching DOI lookups, but I don't think addressing that should hold up this PR, since we don't really have much of a sense of how much overlap there will be. Perhaps just adding a TODO comment would suffice for now?

rialto_airflow/harvest/openalex.py
lwrubel: Added a TODO comment about that.

PR #224, opened 2025-03-07
Connect honeybadger and test

### LINKED ISSUE #205: Investigate HB alerts from airflow tasks
When exceptions occur in airflow tasks, it would be useful to log into HB or similar environment.  Investigate how this could work.

lwrubel:
Resolves #205 to send an HB notification when a task fails. Task-specific callbacks should be set using the DAG's `default_args` rather than being set at the DAG level, so that they're available/called with each task. [Docs](https://airflow.apache.org/docs/apache-airflow/stable/administration-and-deployment/logging-monitoring/callbacks.html). 

Example notification:
https://app.honeybadger.io/projects/130767/faults/118366357


edsu (approved):
Awesome you got this to work!

PR #231, opened 2025-03-07
Include dev_limit in prod docker compose

lwrubel:
No dev_limit is set in puppet for rialto-airflow-prod, just rialto-airflow-stage. 

Goes with https://github.com/sul-dlss/puppet/pull/12026/files


peetucket (approved):
Do we not need this in the compose.yaml ?

lwrubel:
It's handled in local dev situations by putting the value in your `.env` file. So it's only in remote situations where we don't create a .env that we need it.

peetucket:
> It's handled in local dev situations by putting the value in your `.env` file. So it's only in remote (deployed) situations where we don't create a .env that we need it.

ok, i should know this but the compose.prod.yaml is used for both stage and prod vms, and the compose.yaml is for...?

lwrubel:
compose.yaml is what we're using for our local dev. `docker compose up` will by default use a compose.yaml file. In deployed environments, the compose.prod.yaml gets used because of [set :docker_compose_file, 'compose.prod.yaml'](https://github.com/sul-dlss/rialto-airflow/blob/c827cb42c88dc8dec2964a903f390e69b547a9a7/config/deploy.rb#L48)

PR #232, opened 2025-03-07
Dimensions fill in task

### LINKED ISSUE #200: Create "fill-in" task for Dimensions
Similar to #194 we want to query Dimensions by DOI for any publications that lack a `dimensions_json` value, once the initial harvesting by ORCID has finished.

Blocked by #162

lwrubel:
Resolves #200 to fill in Dimensions json where available by DOI. 

This is currently looking up DOIs one by one. With some work it could be adjusted to work with batches of DOIs (and then updating batches of records in the database).

### DIFF
--- a/rialto_airflow/harvest/openalex.py
+++ b/rialto_airflow/harvest/openalex.py
@@ -119,13 +119,12 @@ def fill_in(snapshot: Snapshot, jsonl_file: Path) -> Path:
             )
             # TODO: consider getting data for more than one DOI at a time
             for row in select_session.execute(stmt):
-                logging.info(f"filling in data for {row.doi}")
                 try:
                     openalex_pub = Works()[f"https://doi.org/{row.doi}"]
                     # TODO: get a key so we don't have to sleep!
                     time.sleep(1)
                 except requests.exceptions.HTTPError as e:
-                    logging.error(f"error looking up {row.doi}: {e}")
+                    logging.info(f"No data found for {row.doi}: {e}")
                     continue
 
                 with get_session(snapshot.database_name).begin() as update_session:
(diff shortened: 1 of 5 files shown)


rialto_airflow/harvest/openalex.py
```diff
@@ -119,13 +119,12 @@ def fill_in(snapshot: Snapshot, jsonl_file: Path) -> Path:
...
             # TODO: consider getting data for more than one DOI at a time
             for row in select_session.execute(stmt):
-                logging.info(f"filling in data for {row.doi}")
                 try:
                     openalex_pub = Works()[f"https://doi.org/{row.doi}"]
                     # TODO: get a key so we don't have to sleep!
                     time.sleep(1)
                 except requests.exceptions.HTTPError as e:
-                    logging.error(f"error looking up {row.doi}: {e}")
+                    logging.info(f"No data found for {row.doi}: {e}")
```
lwrubel, line 127: This logging looked like it was going to be excessive in full runs, so removed here as well while working on Dimensions harvesting.

rialto_airflow/harvest/openalex.py
```diff
@@ -119,13 +119,12 @@ def fill_in(snapshot: Snapshot, jsonl_file: Path) -> Path:
             )
             # TODO: consider getting data for more than one DOI at a time
             for row in select_session.execute(stmt):
-                logging.info(f"filling in data for {row.doi}")
```
lwrubel, line 122: This logging looked like it was going to be excessive in full runs, so removed here as well while working on Dimensions harvesting.

rialto_airflow/harvest/openalex.py
lwrubel, line 127: Also tweaked this since its's not an actual error.

jmartin-sul (approved):

PR #237, opened 2025-03-10
Calculate APC

### LINKED ISSUE #236: Store APC charges
Add a step to the Distill task which determines the APC charge related to a publication and stores it in a new `apc` column on the `publication` table.

The logic for doing this is available in the data export [requirements](https://docs.google.com/document/d/1WojtunkzNtidF2JW4ZLcSClbxajMkH36eUdRIl9tk0E/edit?tab=t.0#heading=h.i6i0rin6jumw)

In order of preference we need to:

1. look for the `openalex.apc_paid.value_usd`
1. look in an [external dataset](https://dataverse.harvard.edu/dataset.xhtml?persistentId=doi:10.7910/DVN/CR1MMV) using the `pub_year` (part of #233) and one of the ISSNs
2. look in `openalex.apc_list.value_usd`

Just noting that the external dataset only includes for 2019-2023 -- should we use 2023 numbers for publications from 2024 and 2025?

edsu:
This commit adds an APC extraction based on OpenAlex data and an external dataset which is stored in a new column `apc` in the `publication` table.

https://dataverse.harvard.edu/dataset.xhtml?persistentId=doi:10.7910/DVN/CR1MMV

Closes #236


rialto_airflow/harvest/distill.py
```diff
@@ -103,24 +96,112 @@ def _open_access(pub):
     )
```
edsu, line 98: I moved some functions around, so that similar ones were grouped together.

rialto_airflow/harvest/distill.py
```diff
@@ -45,12 +31,14 @@ def distill(snapshot: Snapshot) -> None:
                 "pub_year": _pub_year(pub),
                 "open_access": _open_access(pub),
                 # These are TBD:
-                # "apc":                _apc(pub),
                 # "funders":            _funders(pub),
                 # "federally_funded":   _federally_funded(pub)
             }
```
edsu, line 37: The APC calculation requires a pub_year, which needs to be extracted first.

rialto_airflow/harvest/distill.py
```diff
@@ -103,24 +96,112 @@ def _open_access(pub):
...
+    jsonp = jsonpath_ng.parse("open_access[*]")
+    for oa in jsonp.find(dim_json):
+        if oa.value and oa.value != "oa_all":
+            return oa.value
+
+    return None
+
+
+def _apc_oa_dataset(openalex, context):
+    if openalex is None:
```
lwrubel, line 143: Looks like you were able to switch this to work with OpenAlex since there are ISSNs there after all?

lwrubel (approved):

rialto_airflow/harvest/distill.py
edsu: Oh whoops, I just realized it's supposed to use the Dimensions ISSN not OpenAlex. I guess it should try to use both? I'll add an issue to take another look at this depending on what @jacobthill would like.

PR #250, opened 2025-03-11
Extract and store publication funders

### LINKED ISSUE #235: Add "extract funders" harvest DAG task
The OpenAccess reporting requirements indicate that we need report on whether a publication is federally funded (boolean) and who the funders are:

https://docs.google.com/document/d/1WojtunkzNtidF2JW4ZLcSClbxajMkH36eUdRIl9tk0E/edit?tab=t.0#heading=h.i6i0rin6jumw

We could add this step to the Distill Task or add it as a new DAG Task following the Distill task (which may need a more specific new name). There's a slight advantage to adding it to the Distill task that is walking through the Publication table, so that we only take one pass. But it might also make sense to do in a targeted way for rows that contain funder data.

The proposed approach in the requirements is to extract funders from the Dimensions JSON data: `.funders` which is a list that looks like:

```json
[
  {
    "id": "grid.419635.c",
    "name": "National Institute of Diabetes and Digestive and Kidney Diseases",
    "types": [
      "Government"
    ],
    "acronym": "NIDDK",
    "linkout": [
      "http://www.niddk.nih.gov/"
    ],
    "latitude": 39.003628,
    "city_name": "Bethesda",
    "longitude": -77.10107,
    "state_name": "Maryland",
    "country_code": "US",
    "country_name": "United States"
  },
  {
    "id": "grid.279885.9",
    "name": "National Heart Lung and Blood Institute",
    "types": [
      "Government"
    ],
    "acronym": "NHLBI",
    "linkout": [
      "http://www.nhlbi.nih.gov/"
    ],
    "latitude": 39.00428,
    "city_name": "Bethesda",
    "longitude": -77.100945,
    "state_name": "Maryland",
    "country_code": "US",
    "country_name": "United States"
  }
]
```

If the Dimensions data isn't available then we are to fall back to looking at the OpenAlex `.grants` information, which looks like:

```json
[
  {
    "funder": "https://openalex.org/F4320306156",
    "award_id": null,
    "funder_display_name": "Children's Cancer Research Fund"
  },
  {
    "funder": "https://openalex.org/F4320337351",
    "award_id": "U24CA55727",
    "funder_display_name": "National Cancer Institute"
  },
  {
    "funder": "https://openalex.org/F4320337351",
    "award_id": "1R01CA78960",
    "funder_display_name": "National Cancer Institute"
  },
  {
    "funder": "https://openalex.org/F4320337351",
    "award_id": "5U10 CA07306",
    "funder_display_name": "National Cancer Institute"
  }
]
```

As you can see this metadata doesn't seem to indicate if the funder is "federal". If you resolve the OpenAlex funder URI you can additional information, that may help, or not. There is additional information in the requirements about using an external CSV to determine whether they are federally funded or not using a GRID ID.

For the reporting we minimally need to know:

- name of the organization
- whether it is a "federal" organization

Since there can be more than one funder for a publication I think we need a many-to-many mapping between Publication and Funder. Ideally we would record relevant identifiers for the organization to help with deduplication.

edsu:
- [x] Create funder database table and mapping from publication
- [x] Move funders.csv into the project
- [x] Create functions to see if a funder name or grid_id is federal
- [x] Store publication funders in the database (inserting into Funder as needed)
- [x] Add publication/funder linking to the Harvest DAG

I didn't seem any errors when running in stage, but we will want to be on the lookout for memory errors with a larger run.

I opted to leave using OpenAlex data for a later PR if we can get more clarity around how we want to do it. I was reluctant to do more than what we did in the previous Jupyter notebook which linked using the GRID ID. I think we want to talk about how we can use OpenAlex's ROR ID instead of assuming names are unique.

Closes #235


rialto_airflow/dags/harvest.py
```diff
@@ -123,9 +124,18 @@ def distill_publications(snapshot, openalex_fill_in, dimensions_fill_in):
         """
         Distill the publication metadata into publication table columns.
         """
-        distill.distill(snapshot)
+        count = distill.distill(snapshot)
```
edsu, line 127: We currently need to return something, in order to get the tasks dependencies to wire up correctly. But I thought it might make more sense to return the number of changes made instead of the snapshot.

rialto_airflow/dags/harvest.py
```diff
@@ -123,9 +124,18 @@ def distill_publications(snapshot, openalex_fill_in, dimensions_fill_in):
...
 
-        return snapshot
+        return count
```
edsu, line 129: This is the new task that will iteratively look at all publications with dimensions data to link them to the funder table.

rialto_airflow/database.py
```diff
@@ -78,6 +78,14 @@ def pg_utcnow(element, compiler, **kw):
 )
```
edsu, line 80: This is the mapping table between Publication and Funder.

rialto_airflow/database.py
```diff
@@ -123,6 +134,21 @@ class Author(Base):  # type: ignore
     )
```
edsu, line 136: This is the new Funder table.

rialto_airflow/funders/239_US_Federal_Funders_for_filter.csv
edsu, line 1: This is the dataset of federal funders linked from our requirements doc. There is a README.md included that indicates where it came from.

rialto_airflow/funders/dataset.py
edsu, line 1: This module provides some functions for determining if a given funder is federal or not. The functionality for looking up funders by name aren't used yet, but are tested. maybe they should be removed for now...

rialto_airflow/funders/linker.py
edsu, line 1: This module provides a function for linking all publications with Dimensions metadata to the Funder table.

rialto_airflow/funders/README.md
```diff
@@ -0,0 +1,4 @@
+The 239_US_Federal_Funders_for_filter.csv dataset was acquired from:
+
+https://dataverse.harvard.edu/dataset.xhtml?persistentId=doi:10.7910/DVN/CR1MMV&version=1.0
```
lwrubel, line 3: Is this the correct URL? I'm getting the APC dataset here.

rialto_airflow/funders/linker.py
```diff
@@ -0,0 +1,56 @@
...
+        for row in session.execute(stmt):
+            count += 1
+            if count % 100 == 0:
+                logging.info(f"processed {count} publications")
+
+            pub = row[0]
+
+            funders = pub.dim_json.get("funders", [])
+            if funders is None:
+                logging.warning(f"got null funders for {pub.dim_json}")
```
lwrubel, line 30: Sound like we don't anticipate a lot of these? Was wondering if logging the full dimensions JSON would make the logs grow too large on a full run. Can also wait and see.

edsu: Yeah, good thought. I'll remove the JSON output.

lwrubel (approved):

Issue #254, opened 2025-03-12
Dimensions 408 timeout error when querying by orcid


lwrubel:
In two runs of the harvest DAG on prod, the dimensions_harvest task failed at the point that the same ORCID was being looked up:

```[2025-03-12, 13:23:02 UTC] {dimensions.py:83} INFO - looking up dois for orcid 0000-0001-6711-3286```

 ```
 File "/opt/airflow/rialto_airflow/dags/harvest.py", line 70, in dimensions_harvest
    return dimensions.harvest(snapshot, limit=harvest_limit)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/airflow/rialto_airflow/harvest/dimensions.py", line 45, in harvest
    for dimensions_pub_json in orcid_publications(orcid_query_str):
  File "/opt/airflow/rialto_airflow/harvest/dimensions.py", line 79, in orcid_publications
    yield from publications_from_dois(dois)
  File "/opt/airflow/rialto_airflow/harvest/dimensions.py", line 114, in publications_from_dois
    result = query_with_retry(q, retry=5)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/airflow/rialto_airflow/harvest/dimensions.py", line 171, in query_with_retry
    raise e
  File "/opt/airflow/rialto_airflow/harvest/dimensions.py", line 165, in query_with_retry
    return dsl().query(q, retry=retry)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/airflow/.local/lib/python3.12/site-packages/dimcli/core/api.py", line 187, in query
    return self.query(
           ^^^^^^^^^^^
  File "/home/airflow/.local/lib/python3.12/site-packages/dimcli/core/api.py", line 187, in query
    return self.query(
           ^^^^^^^^^^^
  File "/home/airflow/.local/lib/python3.12/site-packages/dimcli/core/api.py", line 187, in query
    return self.query(
           ^^^^^^^^^^^
  [Previous line repeated 2 more times]
  File "/home/airflow/.local/lib/python3.12/site-packages/dimcli/core/api.py", line 196, in query
    response.raise_for_status()
  File "/home/airflow/.local/lib/python3.12/site-packages/requests/models.py", line 1024, in raise_for_status
    raise HTTPError(http_error_msg, response=self)
[...]

PR #262, opened 2025-03-13
use dimensions API key instead of user/password

peetucket:
Testing this alternative method to authenticating with Dimensions to see if it avoids the 408 errors in #254 

Wait on https://github.com/sul-dlss/puppet/pull/12056, then deploy branch to stage and test?

Tested in localhost on the command line to verify it works for at least one query (by adding DIMENSIONS_API_KEY setting to local .env file using an API key I minted with my Dimensions account).  Already put key into vault and github secrets, but requires this puppet PR: https://github.com/sul-dlss/puppet/pull/12056 before deploying to ensure the key is available on the VMs

 If this works, there will be a follow-on puppet PR to remove the DIMENSIONS_API_USER and DIMENSIONS_API_PASS values (which will also be removed from vault).  See https://github.com/sul-dlss/puppet/pull/12062

If this doesn't work, we may revert the puppet PR and remove the API key from vault, as well as remove the DIMENSIONS_API_KEY secret from Github.


test/harvest/test_dimensions.py
```diff
@@ -13,9 +12,6 @@
 
 dotenv.load_dotenv()
 
-dimensions_user = os.environ.get("AIRFLOW_VAR_DIMENSIONS_API_USER")
-dimensions_password = os.environ.get("AIRFLOW_VAR_DIMENSIONS_API_PASS")
-
```
peetucket, line 18: not used in tests?

edsu: We do use it in tests.

peetucket: i didn't see it referenced anywhere and tests pass without it -- i think the client is configured via env variables?

jmartin-sul: yeah, looks like the login function itself pulls from the env vars directly in the `dimcli.login` call

jmartin-sul (approved):

jmartin-sul:
https://github.com/sul-dlss/puppet/pull/12056 got merged, so prob fine to merge this whenever?

peetucket:
> [sul-dlss/puppet#12056](https://github.com/sul-dlss/puppet/pull/12056) got merged, so prob fine to merge this whenever?

thanks - yes, i was going to try and test on stage first to see if it made a difference

PR #263, opened 2025-03-13
Publish Open Access CSV

### LINKED ISSUE #215: Write publish step for open access report/dashboard
See notes: https://docs.google.com/document/d/12OJArzr-RPWfMJ042SYTBn6QrPXGbZPCb3KsCUHXAQA/edit?tab=t.0#heading=h.2ehm4pmzpsxp

See data export requirements: https://docs.google.com/document/d/1WojtunkzNtidF2JW4ZLcSClbxajMkH36eUdRIl9tk0E/edit?tab=t.0#heading=h.i6i0rin6jumw

Blocked by #162 and #216 

Will be a new "publish" task in the harvest dag that runs at the end.  There was a previously removed publish task that may not be helpful to review since this is very different.

- [ ] create csv file
- [ ] place in google drive

edsu:
This PR adds a `publish_openaccess` DAG task which writes the `contributions.csv` and `publications.csv` files following the `distill_publications` and `link_funders` tasks. The CSV files are written to the snapshot path.

Publishing to Google Drive will be done separately, but it should be possible to add to the `publish_openaccess` task.

Sample CSV files are attached here: [oa.zip](https://github.com/user-attachments/files/19289215/oa.zip)

Closes #215

PS. all the `type: ignore` statements for mypy / sqlalechemy were a bit unsatisfying, but I don't think they can be avoided until we are on the latest sqlalchemy and Airflow 3?

### DIFF
--- a/rialto_airflow/dags/harvest.py
+++ b/rialto_airflow/dags/harvest.py
@@ -9,6 +9,7 @@ from honeybadger import honeybadger  # type: ignore
 
 from rialto_airflow import funders
 from rialto_airflow.harvest import authors, dimensions, openalex, sul_pub, wos, distill
+from rialto_airflow.publish import openaccess
 from rialto_airflow.database import create_database, create_schema
 from rialto_airflow.snapshot import Snapshot
 from rialto_airflow.utils import rialto_authors_file
@@ -156,6 +157,11 @@ def harvest():
 
         return count
 
+    @task()
+    def publish_openaccess(snapshot):
+        openaccess.write_publications(snapshot)
+        openaccess.write_contributions(snapshot)
+
     snapshot = setup()
 
     snapshot = load_authors(snapshot)
@@ -176,9 +182,13 @@ def harvest():
         snapshot, openalex_jsonl, dimensions_jsonl, wos_jsonl
     )
 
-    distill_publications(snapshot, openalex_fill_in, dimensions_fill_in)
+    distilled_pubs = distill_publications(
+        snapshot, openalex_fill_in, dimensions_fill_in
+    )
 
-    link_funders(snapshot, openalex_fill_in, dimensions_fill_in)
+    linked_pubs = link_funders(snapshot, openalex_fill_in, dimensions_fill_in)
+
+    (distilled_pubs, linked_pubs) >> publish_openaccess(snapshot)
 
 
 harvest()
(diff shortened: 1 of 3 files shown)


rialto_airflow/dags/harvest.py
```diff
@@ -176,9 +182,13 @@ def link_funders(snapshot, openalex_fill_in, dimensions_fill_in):
...
 
-    distill_publications(snapshot, openalex_fill_in, dimensions_fill_in)
+    distilled_pubs = distill_publications(
+        snapshot, openalex_fill_in, dimensions_fill_in
+    )
+
+    linked_pubs = link_funders(snapshot, openalex_fill_in, dimensions_fill_in)
 
-    link_funders(snapshot, openalex_fill_in, dimensions_fill_in)
+    (distilled_pubs, linked_pubs) >> publish_openaccess(snapshot)
```
edsu, line 191: Once the DAG settles a bit I'd like to rewire up the tasks using `>>` where possible, so we don't have to pass around unneeded dependencies?

rialto_airflow/publish/openaccess.py
```diff
@@ -0,0 +1,159 @@
...
+        "federally_funded",
+    ]
+
+    csv_path = snapshot.path / "publications.csv"
+    logging.info(f"started writing publications {csv_path}")
+
+    with csv_path.open("w") as output:
+        csv_output = DictWriter(output, fieldnames=col_names)
+        csv_output.writeheader()
+
```
edsu, line 30: Perhaps it would be helpful to have a comment explaining what this query is doing? I wasn't sure how readable it was.

jacobthill (approved):
This looks good to me in terms of the output. It might be a good idea to have someone who is more familiar with the database take a look as well, but I'm approving it.

rialto_airflow/publish/openaccess.py
lwrubel: I'm trying to understand how `func.jsonb_agg_strict(Funder.name).label("funders"),` works and with the `group_by` using type, so if there's anything to explain there, that could be helpful.

edsu: I added some comments to both blocks of SQL. Do they help at all?

lwrubel: Yes, that helps, thanks!

PR #266, opened 2025-03-14
Add RORs to federal funders dataset.

### LINKED ISSUE #265: Add RORs to the OSTP Impact federal dataset
If we have RORs in addition to GRID IDs in the funders table, we can more easily match funders data from OpenAlex (which uses RORs) and avoid a GRID Id lookup in ROR whenever we want to match funders from the two sources. 

Keep credit for the original Zenodo dataset in the repo.

lwrubel:
Fixes #265 to add RORs to the federal funders dataset. These were obtained via the ROR API. Five of the GRID IDs were not found in ROR look-ups by GRID:

* funder.201191 - Combat Casualty Care Research Program
* funder.57700 (ROR found via manual title lookup)
* grid.515271.5 (ROR found via manual title lookup)
* grid.453355.1 (ROR found via manual title lookup)
* grid.513044.7 - Traumatic Brain Injury Center of Excellence


rialto_airflow/funders/239_US_Federal_Funders_for_filter_with_ror.csv
```diff
@@ -0,0 +1,240 @@
+ID,Name,ror
```
edsu, line 1: how about lowercasing all the headers?

edsu (approved):
I left an (ignorable) suggestion.

rialto_airflow/funders/239_US_Federal_Funders_for_filter_with_ror.csv
lwrubel: I guess now that we're not using the original dataset exactly as it was, this would make sense (and be more predictable). Will do.

PR #267, opened 2025-03-14
schedule publish_orcid dag to run weekly

### LINKED ISSUE #175: Finish publish-orcid DAG
We have code to fetch the orcid statistics from the MAIS ORCID API. We also need the publish-orcid DAG to:

- [x] grab the latest authors.csv file and publish it to Google Drive with only active people (should overwrite existing file).  Note that the task that publishes the authors.csv file needs to filter out inactive people before overwriting the file in google drive (either by creating a new csv file and then uploading or having a task generate a separate authors.csv file from rialto-orgs without inactive people).  The file to replace is called "authors.csv" and is here in the prod shared drive: https://drive.google.com/drive/u/1/folders/1Hrtr5Pc_gviZfwng1G7Ve-NzD7GhS25s
- [ ] schedule the DAG to run weekly
- [x] append the orcid user statistics to the spreadsheet in Google Drive
- [x] Determine the appropriate place in Google Drive for the above files to live
- [x] Point the Tableau dashboard to the files in Google Drive

peetucket:
Fixes #175 - the last part of this ticket is to schedule this DAG to run weekly


rialto_airflow/dags/publish_orcid.py
```diff
@@ -26,7 +26,7 @@
 
 
 @dag(
-    # schedule=@weekly,
+    schedule="@weekly",
```
peetucket, line 29: despite the commented out example, it seems you need to quote that shortcut (or you can use full cron syntax) 

see https://airflow.apache.org/docs/apache-airflow/stable/core-concepts/dag-run.html for examples

edsu (approved):

PR #269, opened 2025-03-17
Batch lookup DOIs during fill-in

edsu:
This commit adjusts the fill-in DOI lookups for Dimensions and OpenAlex to look up the DOIs in batch rather than one by one. We are looking up 50 DOIs at a time for Dimensions and 100 at a time for OpenAlex.

I did adjust some of the other tests to use the `snapshot` fixtures, and in a few case move monkeypatching into the pytest function since it made it easier for me to read without the indirection of a mock.

### DIFF
--- a/rialto_airflow/dags/harvest.py
+++ b/rialto_airflow/dags/harvest.py
@@ -87,7 +87,9 @@ def harvest():
         """
         Fetch the data by ORCID from Dimensions.
         """
-        return dimensions.harvest(snapshot, limit=harvest_limit)
+        jsonl_file = dimensions.harvest(snapshot, limit=harvest_limit)
+
+        return jsonl_file
 
     @task()
     def openalex_harvest(snapshot):
--- a/rialto_airflow/harvest/dimensions.py
+++ b/rialto_airflow/harvest/dimensions.py
@@ -42,6 +42,7 @@ def harvest(snapshot: Snapshot, limit: None | int = None) -> Path:
 
                 # dimensions doesn't want scheme and host
                 orcid_query_str = normalize_orcid(author.orcid)
+
                 for dimensions_pub_json in orcid_publications(orcid_query_str):
                     count += 1
                     if limit is not None and count > limit:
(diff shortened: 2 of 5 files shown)


rialto_airflow/dags/harvest.py
```diff
@@ -87,7 +87,9 @@ def dimensions_harvest(snapshot):
         """
         Fetch the data by ORCID from Dimensions.
         """
-        return dimensions.harvest(snapshot, limit=harvest_limit)
+        jsonl_file = dimensions.harvest(snapshot, limit=harvest_limit)
+
+        return jsonl_file
```
edsu, line 92: I found this a bit easier to read.

rialto_airflow/harvest/dimensions.py
```diff
@@ -186,25 +188,28 @@ def fill_in(snapshot: Snapshot, jsonl_file: Path) -> Path:
...
-                    update_stmt = (
-                        update(Publication)  # type: ignore
-                        .where(Publication.doi == row.doi)
-                        .values(dim_json=dimensions_pub)
-                    )
-                    update_session.execute(update_stmt)
-
-                count += 1
-                jsonl_output.write(json.dumps(dimensions_pub) + "\n")
+
```
edsu, line 191: The partitions() method will give us back chunks of results of length equal to the number used in the `yield_per` on the sqlalchemy call.

test/harvest/test_dimensions.py
```diff
@@ -56,9 +56,8 @@ def f(*args, **kwargs):
     monkeypatch.setattr(dimensions, "orcid_publications", f)
```
edsu, line 58: There are a bunch of changes related to using the `snapshot` fixture, which cuts down on boilerplate a little bit.

test/harvest/test_dimensions.py
```diff
@@ -234,35 +229,67 @@ def test_fill_in(
...
+    mock_publication,
+    mock_no_dim_publication,
+    caplog,
+    monkeypatch,
+):
+    caplog.set_level(logging.INFO)
 
-    def f(*args, **kwargs):
-        raise StopIteration
+    # make it look like Dimensions returns no publications by DOI
```
edsu, line 242: I found moving the monkeypatching into the function to be a bit more readable, than keeping track of the various mocks.

peetucket (approved):

PR #276, opened 2025-03-18
use base google drive id to find folders/files by name

### LINKED ISSUE #273: Consider searching for files in google drive by name
Currently we have different google drives for each dashboard, and are replacing csv files in each of these drive folders for each dashboard.

Since we need IDs for each file we are replacing, we end up having a lot of configuration stored in vault/puppet (one value for each file we are replacing).

An alternate strategy could be:
1. Use a single google drive folder for all files, ensuring each filename was unique across dashboards (perhaps using a pre-pended dashboard name onto the file)
2. Configuring that single google drive folder ID in vault/puppet instead of each individual file file
3. Having the code lookup the file id of the file in the google drive folder instead of having it configured (e.g. like this: https://github.com/sul-dlss/rialto-airflow/blob/main/test/test_google.py#L59-L74

OR 

even if we do not have a single shared google drive, we could have one per dashboard, and then lookup file ids ... this would reduce configuration if were uploading multiple files for a given dashboard since we need only a single configuration for a dashboard instead of one per file uploaded

peetucket:
Fixes #273 

Instead of having to configure the file and folder id of each individual file and dashboard we need to write to/replace/update in vault and puppet, we can instead configured a single shared google drive (which we already do) per environment, and then search within this google drive to sub-folders and files we expect to be there.  We can then fetch the IDs of these expected folders/files and use these.

This greatly reduces the amount of configuration neeed, and allows us to continue uploading/updating existing files by simply updating code and not having to keep updating vault/puppet with new config.

The downside is that we rely on folder and filenames in Google Drive matching our code exactly (and we rely on users being careful to not having duplicate filenames in a shared folder, which causes probems looking up file IDs).

I tested these changes in localhost and they worked fine.

If accepted, we can remove these values from puppet (and the corresponding values it is pulling from vault):

```
 AIRFLOW_VAR_ORCID_INTEGRATION_SHEET_ID
 AIRFLOW_VAR_ORCID_AUTHORS_FILE_ID
 AIRFLOW_VAR_OPEN_ACCESS_PUBLICATIONS_FILE_ID
 AIRFLOW_VAR_OPEN_ACCESS_CONTRIBUTIONS_FILE_ID
```


test/test_google.py
```diff
@@ -56,26 +56,8 @@ def google_file_exists(folder_id, filename):
...
-            pageSize=1,
-            includeItemsFromAllDrives=True,
-            supportsAllDrives=True,
-        )
-        .execute()
-    )
-    files = results.get("files", [])
-    return files[0]["id"] if files else None
-
-
```
peetucket, line 76: this method moved to the core google class, and is test below now

rialto_airflow/google.py
```diff
@@ -8,8 +8,48 @@
...
+            spaces="drive",
+            fields="files(id)",
+            pageSize=1,
+            includeItemsFromAllDrives=True,
+            supportsAllDrives=True,
+        )
+        .execute()
+    )
+    files = results.get("files", [])
+    return files[0]["id"] if files else None
```
peetucket, line 52: moved from test class

rialto_airflow/google.py
```diff
@@ -8,8 +8,48 @@
...
+
+# The Google Drive folder ID for the open access dashboard
+def open_access_dashboard_folder_id():
+    return get_file_id(google_drive_id, "open-access-dashboard")
+
+
+# The Google Drive folder ID for the ORCID dashboard
+def orcid_dashboard_folder_id():
+    return get_file_id(google_drive_id, "orcid-dashboard")
+
```
peetucket, line 28: as we add more dashboard folders in the shared google drive, they will get added here, so we can make them available to the dags

rialto_airflow/dags/publish_orcid.py
```diff
@@ -47,11 +48,17 @@ def update_orcid_integration_stats():
         """
         Get current ORCID integration stats from the ORCID integration API and write to file in Google Drive.
         """
+        orcid_integration_sheet_id = google.get_file_id(
+            google.orcid_dashboard_folder_id(), "orcid-integration-stats"
+        )
```
peetucket, line 53: lookup the id of this google sheet by name instead of in configuration

rialto_airflow/dags/publish_orcid.py
```diff
@@ -36,7 +32,12 @@ def update_authors():
         """
         Update the authors.csv file in Google Drive with the latest authors data CSV file.
         """
-        replace_file_in_google_drive(
+        authors_filename = "authors.csv"
+        orcid_authors_file_id = google.get_file_id(
+            google.orcid_dashboard_folder_id(), authors_filename
+        )
+        logging.info(f"Uploading {authors_filename} to file id {orcid_authors_file_id}")
+        google.replace_file_in_google_drive(
```
peetucket, line 40: lookup the file id of the authors.csv file by name instead of in configuration

rialto_airflow/dags/harvest.py
```diff
@@ -169,20 +167,29 @@ def publish_openaccess(snapshot):
...
-        logging.info(
-            f"Uploading {snapshot.path / 'publications.csv'} to file id {Variable.get('open_access_publications_file_id')}"
+        open_access_publications_filename = "publications.csv"
+        open_access_contributions_filename = "contributions.csv"
+        open_access_publications_file_id = google.get_file_id(
+            google.open_access_dashboard_folder_id(), open_access_publications_filename
+        )
+        open_access_contributions_file_id = google.get_file_id(
+            google.open_access_dashboard_folder_id(), open_access_contributions_filename
         )
```
peetucket, line 177: lookup the file ids of the files to replace by name instead of using configuration

rialto_airflow/google.py
```diff
@@ -8,8 +8,48 @@
 from airflow.providers.google.suite.hooks.sheets import GSheetsHook
 from airflow.providers.google.suite.hooks.drive import GoogleDriveHook
 from airflow.models import Variable
+import dotenv
+
+# Load environment variables
+dotenv.load_dotenv()
 
 gcp_conn_id = Variable.get("google_connection", "google_cloud_default")
+google_drive_id = Variable.get("google_drive_id", "some_google_drive_id")
```
peetucket, line 17: when running in stage/prod/localhost, this value should exist in .env or come from vault
for the tests, a bogus default is fine

test/test_google.py
```diff
@@ -90,6 +72,30 @@ def get_csv_file_contents(file_id):
...
+
+
+def orcid_dashboard_folder_id():
+    folder_id = google.orcid_dashboard_folder_id()
+    assert folder_id is not None
+
+
+def test_get_file_id():
+    # Confirm the file exists in the shared google drive
+    assert google_file_exists(google_drive_id(), "Test") is True
```
edsu, line 87: How o we know there is going to be a file "Test" in the Google Drive folder?

rialto_airflow/dags/harvest.py
edsu: I have a slight preference of using one folder rather than different folders for each dashboard, as I expect there might be some reuse amongst visualizations. But I can live with it being this way.

edsu (approved):

edsu (approved):

test/test_google.py
peetucket: It was in the README, but it needed some extra clarification, which I added (since for this particular test, it's now important that this file be called "Test" and not just be configured with it's ID)

PR #287, opened 2025-03-20
deployment tweaks so that new dependencies are automatically picked up

### LINKED ISSUE #257: capistrano deploy not pulling latest docker image
In the process of upgrading dimcli to v1.4 in the `uv.lock` I built & pushed a new Docker image to Dockerhub:

```shell
$ DOCKER_DEFAULT_PLATFORM="linux/amd64" docker build . -t suldlss/rialto-airflow:latest
$ docker push suldlss/rialto-airflow
```

However I could see from the harvest-dimensions task log that v1.3 was still be used. This prompted me to connect to one of the Airflow Docker containers and I noticed that there was still a `requirements.txt` in there, which we haven't used for a while.

After doing a manual `docker pull suldlss/rialto-airflow` on stage, and doing a `cap deploy` again, I saw that my dimcli update had taken effect. This makes me think that our `cap deploy` process isn't pulling the latest Docker image automatically?

If while addressing this it's possible to parameterize the Docker image that is used during deployment that would allow us to test without having to push changes that we are testing to `suldlss/rialto-airflow:latest` which will be used by production.

jmartin-sul:
fixes #257 

This changes the prod `docker compose` config to `build` using `Dockerfile` and the deployed branch's copy of the code, which will package the dependencies for that branch and bake them into the freshly built Docker image.  This makes it easier to deploy a branch that contains dependency updates (whether for testing or production), since the dependencies are baked into the image, and since we don't auto-update the image from CI (since deployment happens to an on-prem VM using capistrano).


Dockerfile
jmartin-sul, line 1: just a bit of whitespace cleanup in this file, only noticed because i was experimenting with some edits to it

README.md
```diff
@@ -168,15 +168,19 @@ Getting errors with dependencies when trying to run the tests that look somethin
 
 Try clearing the virtual env folder and let it rebuild on the next run: `rm -rf .venv`
 
-## Deployment
+## Pushing an image to Docker Hub
```
jmartin-sul, line 171: the docker hub stuff isn't really integral to deployment now

README.md
```diff
@@ -185,6 +189,8 @@ bundle exec cap prod deploy  # prod
 # Note: there is no QA
 ```
 
+By default, deployment will use the deployed branch's code to build a fresh Docker image locally, and will restart `docker compose`, which will cause it to pick up the new image.  If you'd rather use an image from the Docker Hub registry, see comment by the `build` field in the `x-airflow-common` section of `compose.prod.yaml` for more detailed instructions.
```
jmartin-sul, line 192: a bit more explanation since we don't have consistent patterns around this, since we don't often deploy dockerized apps to our on prem VMs.

compose.prod.yaml
```diff
@@ -51,8 +52,12 @@ x-airflow-common:
   # In order to add custom dependencies or upgrade provider packages you can use your extended image.
   # Comment the image line, place your Dockerfile in the directory where you placed the docker-compose.yaml
   # and uncomment the "build" line below, Then run `docker-compose build` to build the images.
-   #image: ${AIRFLOW_IMAGE_NAME:-apache/airflow:2.9.3}
-  image: ${AIRFLOW_IMAGE_NAME:-suldlss/rialto-airflow:latest}
+  build: .
+  # image: ${AIRFLOW_IMAGE_NAME:-suldlss/rialto-airflow:latest} # to use a registry image, uncomment this and comment or remove
```
jmartin-sul, line 56: this felt like the most usable/visible place for this comment, but happy to move or reformat if others think of more readable ways to explain this.

lwrubel: I'd move it above the `build` element. We might not need the further comment about using a registry image since the Airflow boilerplate above also covers this.

jmartin-sul: d'oh, i think i got boilerplate vision and totally ignored that comment right above this change.  remembered it from other airflow compose files and didn't actually read it this time.

i edited the boilerplate comments to be a little more accurate to our current setup, and made the same change in `compose.yaml` since i'd missed that on the first pass.

jmartin-sul:
ok, i think this is ready for another look:
* i made the change suggested by @lwrubel in review
* i did the docker hub section removal @edsu (iirc) suggested in post-standup discussion
* i left a bit of "if you do want to use docker hub" advice in the deployment section
* i added some container rebuild advice to the dependency management section, missed that before

rebased and repushed

jmartin-sul:
i realized after looking at the aforementioned boilerplate that i had inadvertently followed its instructions in this PR 🙂

edsu (approved):

PR #300, opened 2025-03-25
allow google more time to show the files uploaded before checking

peetucket:
*sigh* tests are still a bit flaky because we rely on google having indicated the file is available ... if this continues to be a problem, we may need to get more creative, but this is making it less flaky for me on localhost


edsu (approved):
Yeah, I believe Google Storage is eventually consistent and the API response will come back before the file is available. Waiting a few seconds longer ain't so bad when we have to wait 12 minutes for the speech-to-text test to pass :-)

PR #301, opened 2025-03-25
update MaIS URLs

peetucket:
Part of https://github.com/sul-dlss/dor-services-app/issues/5284

Waiting on testing/switchover


rialto_airflow/mais.py
```diff
@@ -7,19 +6,11 @@
...
-dotenv.load_dotenv()
 
 # Type aliases for clarity
 ORCIDRecord = dict[str, Any]
 ORCIDStats = list[Union[str, int, float]]
 
-BASE_URL = os.environ.get("AIRFLOW_VAR_MAIS_BASE_URL")
-CLIENT_ID = os.environ.get("AIRFLOW_VAR_MAIS_CLIENT_ID")
-CLIENT_SECRET = os.environ.get("AIRFLOW_VAR_MAIS_SECRET")
-
```
peetucket, line 22: we don't actually need to set any of these values here -- we are always passing them in from the dag when they are used

rialto_airflow/mais.py
```diff
@@ -34,22 +25,11 @@ class TokenFetchError(LookupError):
...
-        "client_id": client_id,
-        "client_secret": client_secret,
-    }
+    token_url = f"{base_url}/oauth2/token"
+    data = {"grant_type": "client_credentials"}
 
     try:
-        response = requests.post(token_url, headers=headers, data=data)
+        response = requests.post(token_url, auth=(client_id, client_secret), data=data)
         response.raise_for_status()  # Raise HTTPError for bad responses (4xx or 5xx)
```
peetucket, line 33: this was not working with their new API scheme, and this fixes it ... also, we didn't need to be passing in the client_id and client_secret in two places (header and payload) anyway

rialto_airflow/mais.py
```diff
@@ -106,24 +86,24 @@ def fetch_orcid_users(
             break
 
         params["page"] += 1  # Increment for next page
-        url = f"{base_url}/mais/orcid/v1{next_link}"  # Construct url using next link value
+        url = f"{base_url}/orcid/v1{next_link}"  # Construct url using next link value
```
peetucket, line 89: their new API has different paths now

rialto_airflow/mais.py
```diff
@@ -106,24 +86,24 @@ def fetch_orcid_users(
...
 
 def fetch_orcid_user(access_token: str, base_url: str, user_id: str) -> ORCIDRecord:
     """Fetches a single ORCID user record by ORCID iD or SUNet ID from the MAIS ORCID API."""
     cleaned_id = normalize_orcid(user_id)
-    return get_response(access_token, f"{base_url}/mais/orcid/v1/users/{cleaned_id}")
+    return get_response(access_token, f"{base_url}/orcid/v1/users/{cleaned_id}")
 
 
-def current_orcid_users(access_token: str) -> list[ORCIDRecord]:
+def current_orcid_users(access_token: str, base_url: str) -> list[ORCIDRecord]:
```
peetucket, line 100: let's just pass in the base_url here like we do for every other method in this file

edsu:
Did you see this run ok? I see the coverage may have taken a hit?

peetucket:
> Did you see this run ok? I see the coverage may have taken a hit?

I did see it run ok locally with the UAT values.  But I can deploy to rialto-airflow stage and update the vault values in stage only the new keys and verify it works there.

Not sure about the drop in coverage.  I think there is mocking going on in the existing tests for this API.

edsu (approved):

PR #314, opened 2025-03-28
ignore years that are in the future

### LINKED ISSUE #305: Enforce data constraints on pub_year field
The latest DAG run published files to Airflow Prod with erroneous publication years e.g. 2105. This causes the  x axis of our bar charts to go wonky.

Image

We need to make sure that `pub_year` is not greater than the current year. Ideally we can update our logic to the following:

if `dimensions.year` <= `current_year`, then `dimensions.year`
elseif `openalex.publication_year` <= `current_year`, then `openalex.publication_year`
elseif sul_pub.year <= `current_year`, then `sul_pub.year`
else `Null`

peetucket:
Fixes #305 by altering the rule for "is_int" (currently only used by years anyway) so that it instead verifies both the fact the year is an int and earlier than the current year.

If we need to validate ints for other fields, we can add that rule separately back in.

Also, fix deprecation warnings.

### DIFF
--- a/rialto_airflow/harvest/distill.py
+++ b/rialto_airflow/harvest/distill.py
@@ -210,9 +210,12 @@ def _first(pub, rules: Rules) -> Optional[str | int]:
                         if year_val <= datetime.datetime.now().year:
                             return year_val
                         else:
-                            logging.warn(f"got a year {year_val} that is in the future")
+                            logging.warning(
+                                f"got a year {year_val} that is in the future"
+                            )
                     except (ValueError, TypeError):
-                        logging.warn(f'got "{value}" instead of int')
+                        # continue matching if the rule wants an int but we don't have one
+                        logging.warning(f'got "{value}" instead of int')
                 else:
                     return value
 
--- a/rialto_airflow/harvest/distill.py
+++ b/rialto_airflow/harvest/distill.py
@@ -8,6 +8,7 @@ from sqlalchemy import select, update
 from rialto_airflow.database import Publication, get_session
 from rialto_airflow.snapshot import Snapshot
 from rialto_airflow.apc import get_apc
+import datetime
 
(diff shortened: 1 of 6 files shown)


rialto_airflow/harvest/distill.py
```diff
@@ -203,12 +204,18 @@ def _first(pub, rules: Rules) -> Optional[str | int]:
             results = jpath.find(data)
             if len(results) > 0:
                 value = results[0].value
-                if rule.is_int:
+                if rule.is_valid_year:
                     try:
-                        return int(value)
+                        year_val = int(value)
+                        if year_val <= datetime.datetime.now().year:
```
lwrubel, line 210: Maybe a question for @jacobthill: would we want to loosen this rule just a little to allow one year beyond the current? 2026 publications will likely be showing up in our sources in late 2025. We would not get them for our reports until we run a harvest in 2026. This depends on what you want to have available in the dashboards.

lwrubel (approved):
Approved, but left a question about the intent of the rule.

rialto_airflow/harvest/distill.py
edsu: There is currently logic for determining the pub_year: 

https://github.com/sul-dlss/rialto-airflow/blob/c847a589a70e2f995c56792706e9428969cee4c4/rialto_airflow/harvest/distill.py#L73-L87

Basically to look in sul_pub, dimensions, openalex, and then web-of-science. This PR would mean that it would check to make sure the pub_year is less than or equal to the current year, and would be applied in that order as well. So if sul_pub has 2026 and dimensions has 2025, we will use 2025.

One advantage to applying the limit at query time is that we can change it easily without needing to reharvest. We could also get a picture of what the publication years look like.

Issue #207, opened 2025-03-04
Web of Science API can return empty string


edsu:
It appears that an API request can return an empty string instead of a JSON object.

Get the Web of Science API Key: 

```shell
$ vault kv get puppet/application/sul_pub/prod/wos-api-key
```

Then try this request for ORCID 280000-0002-3271-7861:

```shell
$ curl --header "application/json" --header 'X-ApiKey: PUT_KEY_HERE' 'https://wos-api.clarivate.com/api/wos?databaseId=WOK&usrQuery=AI%3D%280000-0002-3271-7861%29&count=100&firstRecord=1'
```

And notice you get no output, but that it's a 200 OK. Once we figure out a reason for this we should adjust the `rialto_airflow.harvester.wos`.

We talked about emailing our contact at Clarivate to see if there is a reason that we are seeing this behavior. Otherwise this API route works fine.

edsu:
Noting that @peetucket reports the curl command works for him. It looks like `wos-api.clarivate.com` resolves to multiple IP addresses so maybe I'm talking to one that works a bit differently?

```
$ host wos-api.clarivate.com
wos-api.clarivate.com is an alias for wos-api.prod1.prod.oneplatform.build.
wos-api.prod1.prod.oneplatform.build is an alias for api-pub.us-west-2.prod1.prod.oneplatform.build.
api-pub.us-west-2.prod1.prod.oneplatform.build is an alias for kongapi-prod-us-west-2.prod.sp.aws.clarivate.net.
kongapi-prod-us-west-2.prod.sp.aws.clarivate.net has address 44.225.8.221
kongapi-prod-us-west-2.prod.sp.aws.clarivate.net has address 35.155.244.135
kongapi-prod-us-west-2.prod.sp.aws.clarivate.net has address 44.228.82.11
```

Adding `--verbose` to the `curl` command highlights that I'm talking to `44.225.8.221`:

```
* Host wos-api.clarivate.com:443 was resolved.
* IPv6: (none)
* IPv4: 44.225.8.221, 35.155.244.135, 44.228.82.11
*   Trying 44.225.8.221:443...
* Connected to wos-api.clarivate.com (44.225.8.221) port 443
* ```

edsu:
I think there was some confusion about the request in question, and that this one consistently fails with an empty string response.

I think we discussed contacting Clarivate to see if there is a fix on their side so that we can collect publications for these authors?

peetucket:
[...]

PR #328, opened 2025-04-01
issue SR request for WoS APIs calls to limit data returned

peetucket:
To deal with #207 

Just added the WoS SR param without investigating too deeply what fields are missing and if we use them.  I then ran a full harvest of 100 records on my laptop, whocj ran successfully, so it's likely we were never using the fields that aren't returned anyway.

Ran harvest of 100 on localhost = successfully completed
Ran harvest of 1000 on stage = successfully completed


lwrubel:
@jacobthill FYI since could affect data coverage analysis of fields, if this harvested data were used.

lwrubel (approved):

Issue #330, opened 2025-04-03
Update APC values based on Open Access status


jacobthill:
Take apc column, overwrite publications where open_access=”Closed” (non OA publications should always be $0) and fill empty cells based on the following formula:

- if open_access=”Diamond”, value should be $0
- elseif open_access=”Gold”, value should be $2,450
- elseif open_access=”Hybrid”, value should be $3,600

PR #331, opened 2025-04-03
Set APCs for closed pubs to 0

lwrubel:
Resolves the first requirement of #330, to set APCs for "closed" publications to 0, since a closed article should not have an estimated APC.

### DIFF
--- a/rialto_airflow/harvest/distill.py
+++ b/rialto_airflow/harvest/distill.py
@@ -35,7 +35,7 @@ def distill(snapshot: Snapshot) -> int:
                 "open_access": _open_access(pub),
             }
 
-            # pub_year in cols is needed to determine the apc
+            # pub_year and open_access in cols is needed to determine the apc
             cols["apc"] = _apc(pub, cols)
 
             # update the publication with the new columns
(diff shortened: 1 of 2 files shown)


lwrubel:
Tests are just failing with `FAILED test/harvest/test_dimensions.py::test_408_errors - Failed: DID NOT RAISE <class 'requests.exceptions.HTTPError'>` which is an indicator Dimensions may have fixed the issue with problem set of DOIs. Unrelated to this work.

peetucket (approved):

peetucket:
> Tests are just failing with `FAILED test/harvest/test_dimensions.py::test_408_errors - Failed: DID NOT RAISE <class 'requests.exceptions.HTTPError'>` which is an indicator Dimensions may have fixed the issue with problem set of DOIs. Unrelated to this work.

I've noticed this is flaky - it sometimes passes and sometimes fails.   not sure what this is telling us

edsu:
I've not seen this test fail before, interesting. 

If we aren't going to address the entire issue with this PR lets merge this as is, and keep the issue open?

lwrubel:
408 errors change ticketed here: https://github.com/sul-dlss/rialto-airflow/issues/333

PR #332, opened 2025-04-03
add faculty column to open access publish step

### LINKED ISSUE #326: Add faculty_authored column to Open Access data export
It should be a boolean column with True if any Stanford author has the role of “faculty”, otherwise False. This is analogous to the `academic_council` field.

peetucket:
Fixes #326 

(note that the existing column for academic council is not called `academic_council_authored` as mentioned in the ticket, it is just `academic_council`, not sure if this is correct or needs to be changed)


peetucket:
Interesting, the dimensions 408 test is failing in this PR too

lwrubel:
Issue to stop testing for 408 errors and try restoring the researchers field noted here: https://github.com/sul-dlss/rialto-airflow/issues/333.

lwrubel:
This adds the `faculty_authored` column to `contributions.csv`. The way I read the ticket, @jacobthill wanted this in the `publications.csv`, with logic to look at all of the authors associated with the publication to determine if any one of them has primary_role of "faculty". 

We don't currently include any author-related info in the `publications.csv` so the authors data will need to be joined with the publication in the query.

test/publish/test_openaccess.py
```diff
@@ -110,6 +110,7 @@ def test_write_contributions(test_session, snapshot, dataset, caplog):
     assert row.sunet == "janes"
     assert row.role == "faculty"
     assert bool(row.academic_council) is True
+    assert bool(row.faculty_authored) is True
```
edsu, line 113: This is a bit of a nit pick, but could we have `faculty` instead of `faculty_authored`. We don't have `academic_council_authored` or `role_authored`, etc...

edsu (changes requested):

test/publish/test_openaccess.py
jacobthill: Actually we do have an `academic_council_authored` ticket. We don't need a `role_authored` column because we don't use that in our charts. But my concern with `faculty` vs `faculty_authored` is its easy to get confused with what the data actually means. `faculty_authored` is clear to me and the name implies that the data is about the publication. `academic_council` is another column that is connected to a person (is that person a member of the academic council). If we name it `faculty` it could lead us to falsely assume that a `TRUE` value in that column means the author is a faculty member. We can talk through this at standup if need be.

edsu: Ok, consistency is important, so perhaps this PR could address those other naming issues?

edsu (approved):
Approving, whether naming issues are addressed here, or documented in a new ticket.

peetucket:
closed in favor of #334

PR #334, opened 2025-04-04
Add publications.csv columns

### LINKED ISSUE #293: Add academic_council_authored field to open access publications.csv file
We need a boolean field called `academic_council_authored` that is True if any of the Stanford authors are members of the academic council, otherwise its False.

edsu:
Adds two columns to the publications.csv

- academic_council_authored
- faculty_authored

closes #293
closes #332

### DIFF
--- a/rialto_airflow/publish/openaccess.py
+++ b/rialto_airflow/publish/openaccess.py
@@ -19,6 +19,8 @@ def write_publications(snapshot) -> Path:
         "types",
         "funders",
         "federally_funded",
+        "academic_council_authored",
+        "faculty_authored",
     ]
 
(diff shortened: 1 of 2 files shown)


rialto_airflow/publish/openaccess.py
```diff
@@ -61,9 +68,11 @@ def write_publications(snapshot) -> Path:
                         "pub_year": row.pub_year,
                         "apc": row.apc,
                         "open_access": row.open_access,
-                        "types": "|".join(_get_types(row)) or None,
-                        "funders": "|".join(row.funders) or None,
```
edsu, line 65: It helps for the values to be sorted when testing.

PR #341, opened 2025-04-14
Add ROR ID from OpenAlex to Funder

### LINKED ISSUE #324: Start adding RORs to funder records
Since OpenAlex funder records usually have RORs, start adding them to funders in the database. There is already a ror_id field in the funder model. Note that this needs to be accounted for when adding funders in the linker task (see TODOs) and check for conflicts.

### LINKED ISSUE #338: Optimize link_funders
Prior to https://github.com/sul-dlss/rialto-airflow/pull/323 the link_funders task took 6 hours. After it was merged it is still running after 42 hours. We should see where it is spending its time and see if it could be sped up.

This may be unrelated but I was expecting to see some rows in the `funder` table where `ror_id` as not null. But even after running for this time that column is still empty.

edsu:
Ensure that the ROR ID that is obtained from OpenAlex is persisted to Funder. OpenAlexID is saved too which means we don't need to query OpenAlex multiple times for the same OpenAlexID. Closes #324

The logic for inserting into `pub_funder_association` and `funder` was updated to share a session after I noticed a slowdown was caused by a database deadlock, that I believe was caused by similar edits in different sessions (but I'm not entirely sure):

```
rialto_20250412204405=# SELECT client_addr, client_port, query                                                          FROM pg_stat_activity                                                                                                   WHERE state =                                                                                                           'active'                                                                                                    AND         wait_event_type = 'Lock'                                                                                          ;

client_addr | 172.19.0.7
client_port | 58004
query       | INSERT INTO funder (name, grid_id, ror_id, openalex_id, federal) VALUES ('Department of Medicine,         University of California, San Francisco', 'grid.266102.1', 'https://ror.org/043mz5j54', 'https://openalex.org/          F4320337582', false) ON CONFLICT ON CONSTRAINT funder_grid_id_key DO UPDATE SET name = 'Department of Medicine,         University of California, San Francisco', grid_id = 'grid.266102.1', ror_id = 'https://ror.org/043mz5j54', openalex_id  = 'https://openalex.org/F4320337582', federal = false RETURNING funder.id
```

The row containing `grid.266102.1` already existed in the `funder` table. This query was locked for multiple hours which caused the link_funders task to just stop. The lock happened early on, after processing like 180 OpenAlex funders. After these changes I was able to run it on 180k publications in my development environment.

Closes #338

### DIFF
--- a/rialto_airflow/database.py
+++ b/rialto_airflow/database.py
@@ -141,6 +141,7 @@ class Funder(Base):  # type: ignore
     name = Column(String, nullable=False)
     grid_id = Column(String, unique=True)
     ror_id = Column(String, unique=True)
+    openalex_id = Column(String, unique=True)
     federal = Column(Boolean, default=False)
     created_at = Column(DateTime, server_default=utcnow())
     updated_at = Column(DateTime, onupdate=utcnow())
--- a/rialto_airflow/funders/linker.py
+++ b/rialto_airflow/funders/linker.py
@@ -1,21 +1,22 @@
 import logging
 import os
+import time
 from typing import Optional
 
+import requests
 from pyalex import Funders, config
 from sqlalchemy import select
+from sqlalchemy.orm import Session
 from sqlalchemy.dialects.postgresql import insert
 
-
 from rialto_airflow.database import (
-    get_session,
-    Publication,
     Funder,
+    Publication,
+    get_session,
     pub_funder_association,
 )
 from rialto_airflow.funders.dataset import is_federal, is_federal_grid_id
 from rialto_airflow.funders.ror_grid_dataset import convert_ror_to_grid
-
 
 config.email = os.environ.get("AIRFLOW_VAR_OPENALEX_EMAIL")
 config.max_retries = 5
@@ -55,7 +56,7 @@ def link_dim_publications(snapshot) -> int:
 
             with get_session(snapshot.database_name).begin() as update_session:
                 for funder in pub.dim_json.get("funders", []):
-                    if funder_id := _find_or_create_funder(snapshot, funder):
+                    if funder_id := _find_or_create_dim_funder(update_session, funder):
                         update_session.execute(
                             insert(pub_funder_association)
                             .values(publication_id=pub.id, funder_id=funder_id)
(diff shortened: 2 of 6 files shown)


rialto_airflow/database.py
```diff
@@ -141,6 +141,7 @@ class Funder(Base):  # type: ignore
     name = Column(String, nullable=False)
     grid_id = Column(String, unique=True)
     ror_id = Column(String, unique=True)
+    openalex_id = Column(String, unique=True)
```
edsu, line 144: Saving the OpenAlexID means we can use it to look up a Funder instead of always needing to go to the API.

rialto_airflow/funders/linker.py
```diff
@@ -90,23 +91,20 @@ def link_openalex_publications(snapshot) -> int:
...
+                        .on_conflict_do_nothing()
                     )
-                    if funder_id := _find_or_create_funder(snapshot, openalex_funder):
-                        update_session.execute(
-                            insert(pub_funder_association)
-                            .values(publication_id=pub.id, funder_id=funder_id)
-                            .on_conflict_do_nothing()
-                        )
     logging.info(f"processed {count} publications from OpenAlex")
     return count
```
edsu, line 106: `_find_or_create_funder` was split into `_find_or_create_dim_funder` and `_find_or_create_openalex_funder` since they need to do different things.

rialto_airflow/funders/linker.py
```diff
@@ -116,35 +114,81 @@ def _find_or_create_funder(snapshot, funder: dict) -> Optional[int]:
...
-def _lookup_openalex_funder(funder: dict) -> Optional[dict]:
+def _find_or_create_openalex_funder(
+    session: Session, openalex_id: str
+) -> Optional[int]:
+    # if the funder is in the database aleady return it
+    funder = session.execute(
+        select(Funder).where(Funder.openalex_id == openalex_id)
+    ).scalar_one_or_none()
+    if funder:
+        return funder.id
```
lwrubel, line 137: Would be good to add test coverage for this?

rialto_airflow/funders/linker.py
```diff
@@ -116,35 +114,81 @@ def _find_or_create_funder(snapshot, funder: dict) -> Optional[int]:
...
+        funder = Funders()[openalex_id]
+        logging.info(f"found funder data in openalex for {openalex_id}")
+    except requests.exceptions.HTTPError:
+        logging.warn(f"OpenAlex API returned error for {openalex_id}")
         return None
+
+    # a ror_id is required
+    ror_id = funder.get("ids", {}).get("ror")
+    if ror_id is None:
+        logging.warn(f"no ROR ID for {openalex_id}")
```
lwrubel, line 176: I think this is a normal situation, especially with small funding orgs, so could be logged at the info level.

rialto_airflow/funders/linker.py
```diff
@@ -90,23 +91,20 @@ def link_openalex_publications(snapshot) -> int:
 
         with get_session(snapshot.database_name).begin() as update_session:
             for funder in funders:
```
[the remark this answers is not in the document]
lwrubel: Definitely a helpful refactor.

lwrubel (approved):
Left some small comments, but approved. Glad to know you were able to successfully run with 180K harvest limit and it goes much more quickly and reliably.

rialto_airflow/funders/linker.py
edsu: It might be normal, but currently we are ignoring this Funder information from OpenAlex, which seems not great?

lwrubel: Sure, and it will be good to see how many don't have RORs.

rialto_airflow/funders/linker.py
edsu: Good idea. I added one.

edsu:
This has been merged into main, but for some reason the PR seems to be still open and the check for the ability to merge is just a spinner?

PR #345, opened 2025-04-14
Lookup publications by ORCID

### LINKED ISSUE #322: harvest_dimensions is looking up publications twice
During the `dimensions_harvest` we are looking up publications by ORCID, to get a list of publication DOIs, and then looking up that list of DOIs in Dimensions to get the full metadata for the same publications:

```
[2025-04-01, 00:18:02 UTC] {dimensions.py:79} INFO - looking up publications for orcid 0000-0003-0800-1860
[2025-04-01, 00:18:03 UTC] {dimensions.py:108} INFO - looking up: "10.11588/xxi.2024.4.108255","10.52214/uw.v32i.12541","10.16995/ah.11575","10.1017/s0020743824000084","10.16995/ah.8937","10.4324/9781003170525-7","10.1163/22118993-00361p06","10.1386/ijia_00025_1","10.1080/00210862.2019.1675133","10.1515/9783110563443-006","10.1017/rms.2018.12","10.1080/09503110.2015.1102494","10.1086/692804","10.1007/978-94-007-7747-7_10227","10.1007/978-94-007-3934-5_10227-1","10.1080/13645145.2012.723862"
[2025-04-01, 00:18:09 UTC] {dimensions.py:79} INFO - looking up publications for orcid 0000-0002-4921-7193
```

This logic was left over from our previous way of building a full set of DOIs prior to harvesting.

This can now be simplified (and sped up) by simply looking up the publications by ORCID and saving those to the database, without needing an interstitial list of DOIs. This would involve rewriting `orcid_publications` to work more like `publications_from_dois` and removing `dois_from_orcid`. Perhaps we should rename `orcid_publications` to be `publications_from_orcid`?

We do still need `publications_from_dois()` for the `fillin_dimensions` DAG task.

edsu:
Instead of looking up DOIs by ORCID and then turning around and getting publication metadata for those DOIs we can simply query for publication metadata by ORCID. This should reduce the number of API calls during the initial harvest of metadata. We do still need to be able to lookup publications by DOIs during the fill in process.

In testing with a limit of 100k in my dev environment I can see that the `dimensions_harvest` task went from 3.5 hours to 2 hours.

I also changed the call to dimensions `query()` to using `query_iterative()` which will page through additional results, whereas previously we were limited to 1000.

https://digital-science.github.io/dimcli/modules.html#dimcli.core.api.Dsl.query_iterative

Closes #322

### DIFF
--- a/rialto_airflow/harvest/dimensions.py
+++ b/rialto_airflow/harvest/dimensions.py
@@ -40,10 +40,7 @@ def harvest(snapshot: Snapshot, limit: None | int = None) -> Path:
                     logging.info(f"Reached limit of {limit} publications stopping")
                     break
 
-                # dimensions doesn't want scheme and host
-                orcid_query_str = normalize_orcid(author.orcid)
-
-                for dimensions_pub_json in orcid_publications(orcid_query_str):
+                for dimensions_pub_json in publications_from_orcid(author.orcid):
                     count += 1
                     if limit is not None and count > limit:
                         stop = True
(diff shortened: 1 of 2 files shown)


rialto_airflow/harvest/dimensions.py
```diff
@@ -110,14 +84,31 @@ def publications_from_dois(dois: list, batch_size=200):
         q = f"""
             search publications where doi in [{doi_list}]
             return publications [{fields}]
-            limit 1000
```
edsu, line 113: now that `query_iterative` is being used this limit is not needed.

peetucket (approved):

PR #349, opened 2025-04-15
schedule harvest to run weekly

### LINKED ISSUE #2: Configure the harvest airflow DAG to run weekly
Add the rialto-service@lists.stanford.edu email list to receive reports on DAG runs

Not on a monday so we don't bump into dependency update deploys.  Late Tuesday or Wednesday would be good

peetucket:
1. Schedule harvest dag to run weekly.
2. Do not allow more than one concurrent DAG to be running for either orcid or harvest (so that if the previous run is still happening, a new one will not start).

Fixes #2 

see https://www.astronomer.io/docs/learn/rerunning-dags/

### DIFF
--- a/rialto_airflow/dags/harvest.py
+++ b/rialto_airflow/dags/harvest.py
@@ -58,7 +58,8 @@ def task_failure_notify(context):
 
 
 @dag(
-    schedule=None,
+    schedule="@weekly",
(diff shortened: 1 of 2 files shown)


edsu (approved):

edsu:
nice find on max_active_runs!

PR #350, opened 2025-04-16
Don't query OpenAlex for DOIs containing pmcid:

edsu:
When querying OpenAlex for multiple DOIs, where one of the DOIs contains the substring `pmcid:` we encounter an OpenAlex error:

```
pyalex.api.QueryError: It looks like you're trying to do an OR query between filters and it's not supported.
You can do this: institutions.country_code:fr|en, but not this: institutions.country_code:gb|host_venue.issn:0957-1558.
```

We need to ignore DOIs like this during the fill_in_openalex task. We received a DOI like this from sulpub prod:

sulpub_id: 895752
doi: 10.1093/noajnl/vdad070.013 pmcid: pmc10402389

### DIFF
--- a/rialto_airflow/harvest/openalex.py
+++ b/rialto_airflow/harvest/openalex.py
@@ -121,18 +121,8 @@ def fill_in(snapshot: Snapshot, jsonl_file: Path) -> Path:
                 # since the query uses yield_per=50 we will be looking up 50 DOIs at a time
                 dois = [normalize_doi(row["doi"]) for row in rows]
 
-                # Commas are a reserved character in openalex, so until there is
-                # a way to escape them we will need to drop them
-                # https://docs.openalex.org/how-to-use-the-api/get-lists-of-entities/filter-entity-lists#intersection-and
-                #
-                # Also if a doi starts with 'doi:' that confuses the OpenAlex
-                # API because it interprets it as trying to change the field
-                # that is being matched as trying to do an OR query with
-                # multiple fields.
-
-                dois_filtered = filter(
-                    lambda doi: "," not in doi and not doi.startswith("doi:"), dois
-                )
+                # drop dois that are problematic for the openalex api
+                dois_filtered = _clean_dois_for_query(dois)
 
                 # looking up multiple DOIs is supported by pipe separating them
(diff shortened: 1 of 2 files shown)


peetucket:
How about if we use a library to just check for valid DOI format before even adding them to the list?  This would cover more cases (of which I am sure we will find other junk in sul-pub)

https://github.com/inveniosoftware/idutils?tab=readme-ov-file
https://idutils.readthedocs.io/en/latest/

Something like
```
import idutils
idutils.is_doi('10.1000/xyz123')  # Returns True if valid format
```

edsu:
Thanks Copilot, but no thanks.

Validating DOIs is difficult because it's not simply a matter of comparing a DOI to the current rules, since they are permanent identifiers. Just because the current rules prevent a particular string doesn't mean it isn't a valid DOI.

At any rate, idutils won't help us in these cases with the OpenAlex API because idutils says these are valid, but they cause OpenAlex API to throw an error.

```
>>> idutils.validators.is_doi('10.1093/hmg/,dds161')
<re.Match object; span=(0, 19), match='10.1093/hmg/,dds161'>

>>> idutils.validators.is_doi('10.1093/noajnl/vdad070.013 pmcid: pmc10402389')
<re.Match object; span=(0, 45), match='10.1093/noajnl/vdad070.013 pmcid: pmc10402389'>

>>> idutils.validators.is_doi('doi:10.1093/hmg/dds161')
<re.Match object; span=(0, 22), match='doi:10.1093/hmg/dds161'>
```

The problem isn't whether these are valid DOIs so much as it is an issue with what OpenAlex API will allow.

peetucket:
ok, that idutils is not working as I'd expect, i was thinking it would return false for something like `10.1093/hmg/,dds161` or `10.1093/noajnl/vdad070.013 pmcid: pmc10402389` and then we'd simply ignore at the point we are harvesting, like here: https://github.com/sul-dlss/rialto-airflow/blob/main/rialto_airflow/harvest/sul_pub.py#L107-L111

so basically, bad dois wouldn't even make it through the harvest, they'd be filtered out

but that can be done separately

Issue #318, opened 2025-03-31
Strategy for handling growth


edsu:
We have 500 GB of storage on the sul-rialto-airflow-db-production server for PostgreSQL. The snapshot we have at the moment seems to be about 30GB. Similarly the `/rialto-data` shared volume where we store collected JSON data is 500 GB and each snapshot appears to be about 30 GB.

We can expect both the size of the databases and the size of the shared volume to grow over time as there are more publications, and more sources (e.g. pubmed). 

We need a plan for how to handle this growth. Approaches could include some combination of:

1. Implement logic for purging unneeded snapshots of data on the shared volume and databases.
2. Adjust the Harvest DAG to allow it to *update* a database rather than starting with a new database each time.

jacobthill:
We need a story time.

peetucket:
We now have implemented a cleanup of old files and databases to avoid filling up disk (see #354)  Leaving this open for future consideration of an improved strategy (eg. incremental updates).

peetucket:
From @jmartin-sul:


my only thought on this front, which would of course add complexity: maybe yet another DAG could archive one snapshot per quarter or per year or something, to a space that won't get cleaned up. that way, we have indefinite history, but maybe not as dense and space consuming as it'd be if we kept everything

jmartin-sul:
> From [@jmartin-sul](https://github.com/jmartin-sul):
> 
> my only thought on this front, which would of course add complexity: maybe yet another DAG could archive one snapshot per quarter or per year or something, to a space that won't get cleaned up. that way, we have indefinite history, but maybe not as dense and space consuming as it'd be if we kept everything

ok, one other thought 🙂 -- if instead of making that archiving behavior a separate DAG, we made it a task on the cleanup DAG added in #354, it could be a pre-req for the other cleanup tasks.  then it would have to finish running first, helping to avoid race condition type behavior from cleanup and archiving possibly running concurrently.

peetucket:
suggestions on how we would know when to do an archive?  currently the cleanup logic is pretty simple: older than X days, delete it...

maybe something like:

1. example last archived directory for most recent snapshot in the archive
2. if most recent copy in archive directory greater than Y (e.g. 6 months), copy latest snapshot to archive
3. then proceed with cleanup as normal

snapshot and author files will be easy, as it is just a copy operation

the database might be trickier...maybe a dump to a file and then archive the file?

edsu:
[...]

PR #354, opened 2025-04-16
Purge old data

peetucket:
A first strategy to avoid filling up the disk (see #318).  

Configure a set number of days and when this DAG is run it will remove any author files and snapshot folders older than this number of days.  Downside is: history is lost.

To do:
- [x] tests
- [x] do remove database task

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,7 +78,9 @@ docker compose up -d
 
 ### Console
 
-$ uv run dotenv run python
+```
+uv run dotenv run python
+```
 
 ### Set-up
 
--- a/compose.prod.yaml
+++ b/compose.prod.yaml
@@ -107,6 +107,7 @@ x-airflow-common:
     AIRFLOW_VAR_RIALTO_POSTGRES: "postgresql+psycopg2://${DATABASE_USERNAME}:${DATABASE_PASSWORD}@${DATABASE_HOSTNAME}"
     AIRFLOW_VAR_HONEYBADGER_API_KEY: ${HONEYBADGER_API_KEY}
     AIRFLOW_VAR_HONEYBADGER_ENV: ${HONEYBADGER_ENV}
+    AIRFLOW_VAR_CLEANUP_INTERVAL_DAYS: 30
   volumes:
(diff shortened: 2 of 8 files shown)


edsu (approved):
I guess it's weird to review my own code, but I think this is worth merging once you are happy with it.

I did notice that it's theoretically possible for a the cleanup step to remove the database and snapshot directory for a harvest that's underway. But since the CLEANUP_INTERVAL_DAYS is set to 10 and 30 that seems unlikely to happen given the current amount of time that it takes for the harvest to run.

peetucket:
Note: I added a new test which adds coverage for the new `drop_database` and `database_exists` methods too

lwrubel:
I see that the authors files in the data_dir are named MM-DD-YYYY (e.g. `authors.csv.02-20-2025`). But the code seems to assume they are YYYY-MM-DD. That seems like a problem, unless I'm misunderstanding things.

peetucket:
> I see that the authors files in the data_dir are named MM-DD-YYYY (e.g. `authors.csv.02-20-2025`). But the code seems to assume they are YYYY-MM-DD. That seems like a problem, unless I'm misunderstanding things.

good catch -- we have the data mocked incorrectly 

i will correct this

peetucket:
> > I see that the authors files in the data_dir are named MM-DD-YYYY (e.g. `authors.csv.02-20-2025`). But the code seems to assume they are YYYY-MM-DD. That seems like a problem, unless I'm misunderstanding things.
> 
> good catch -- we have the data mocked incorrectly
> 
> i will correct this

and the code is also wrong of course, to match the wrong mocked data...i will fix that too

lwrubel (approved):
Reviewed the updated code and approved, pending checks passing.

jmartin-sul (approved):

jmartin-sul:
> Configure a set number of days and when this DAG is run it will remove any author files and snapshot folders older than this number of days. Downside is: history is lost.

my only thought on this front, which would of course add complexity: maybe yet another DAG could archive one snapshot per quarter or per year or something, to a space that won't get cleaned up.  that way, we have indefinite history, but maybe not as dense and space consuming as it'd be if we kept everything.

PR #372, opened 2025-04-28
simplify how we call mais api

### LINKED ISSUE #371: The MaiS ORCID client code should refresh the token if it expires
The Mais ORCID API seems to need a new token every 5 minutes.  Our gem code automatically detects an auth failure and tries to fetch a new token, but the rialto-airflow implementation does not do this.

See https://github.com/sul-dlss/mais_orcid_client/blob/main/lib/mais_orcid_client/token_wrapper.rb

peetucket:
1. The MaIS API uses a GET with simple URL params to make request.  We don't need that params hash.  This aligns it's usage with how we do it in the Ruby mais-orcid-client gem.
2. Move token generation into the method that does pagination, so we can more easily fetch a new token if we get a 401.
3. Add some extra logging of what the links are in the response so we can better see what the MaIS API is returning.
4. Fix an issue that was preventing the MaIS tests from running locally when credentials were available.
5. Retry to get a new token if one expires (three times)

Tested in localhost and via `uv run pytest test/test_mais.py` on my laptop where i have credentials for the MaIS API

Fixes #371

### DIFF
--- a/rialto_airflow/mais.py
+++ b/rialto_airflow/mais.py
@@ -65,19 +65,41 @@ def fetch_orcid_users(
     mais_client_secret: str,
     mais_token_url: str,
     base_url: str,
-    path: str,
     limit: Optional[int] = None,
 ) -> list[ORCIDRecord]:
     """Fetches ORCID user records from the MAIS ORCID API, handling pagination."""
 
     orcid_users: list[ORCIDRecord] = []
     per_page = 100  # This max value from the API.
-    url = f"{base_url}/orcid/v1{path}&page_number=1&page_size={per_page}"  # first page url
+    url = f"{base_url}/orcid/v1/users?scope=ANY&page_number=1&page_size={per_page}"  # first page url
     access_token = get_token(mais_client_id, mais_client_secret, mais_token_url)
 
-    while True:
+    while True:  # paging
         logger.info(f"Fetching {url}")
-        data = get_response(access_token, url)
+        tries = 0
+        while True:  # retry loop for token expiration
+            try:
+                # Try to get the response
+                data = get_response(access_token, url)
+                break  # break out of the retry loop if we get a successful response
+            except requests.exceptions.HTTPError as e:
+                # if we get a 401 unathorized error from the API call, try to get a new token
+                if e.response.status_code == 401:
+                    tries += 1
+                    logger.warning(
+                        f"Attempt {tries}: Access token expired. Retrying..."
+                    )
+                    access_token = get_token(
+                        mais_client_id, mais_client_secret, mais_token_url
+                    )
+                    if tries >= 3:
+                        logger.error(
+                            "Failed to fetch new access token after multiple attempts."
+                        )
+                        raise e  # break out of the retry loop if we fail three times to get a new token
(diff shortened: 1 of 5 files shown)


peetucket:
Drop in code coverage is because CI cannot run the MaIS tests (they actually hit MaIS APIs, and can only be run locally)

test/test_mais.py
```diff
@@ -74,27 +76,17 @@ def test_get_response_success(access_token, base_url, client_id, client_secret):
...
-@pytest.mark.mais_tests
-def test_fetch_orcid_users_invalid_path(
-    access_token, base_url, client_id, client_secret
-):
-    if not (client_secret and client_id):
-        pytest.skip("No MAIS credentials available")
-    invalid_path = "/invalid_path"
-    with pytest.raises(requests.exceptions.HTTPError):
-        mais.fetch_orcid_users(access_token, base_url, invalid_path)
-
```
peetucket, line 96: this test is not useful anymore as the path is no longer a parameter

rialto_airflow/mais.py
```diff
@@ -50,43 +50,76 @@ def get_token(client_id: str, client_secret: str, base_url: str) -> str:
...
-    response = oauth_session.get(url, params=params)
+    response = oauth_session.get(url)
     response.raise_for_status()  # Raise HTTPError for bad responses
     return response.json()
 
 
+def page_size() -> int:
+    """Returns the page size for the API request."""
+    return 100  # This is the max value from the API.
+
```
peetucket, line 66: this lets us easily alter the page size in the tests to verify paging works in UAT (which has less than 100 users)

rialto_airflow/mais.py
```diff
@@ -50,43 +50,76 @@ def get_token(client_id: str, client_secret: str, base_url: str) -> str:
...
 
         if limit and len(orcid_users) >= limit:
             break
 
-        next_link = data.get("links", {}).get("next")
-        if not next_link:
+        self_link = data.get("links", {}).get("self")
+        last_link = data.get("links", {}).get("last")
+        # if we have reached the last page, the self and last links will be the same
+        if self_link == last_link:
```
edsu, line 117: I think this is fine, or it could be simplified by just stopping when there is no next_link?

test/test_mais.py
```diff
@@ -74,27 +76,17 @@ def test_get_response_success(access_token, base_url, client_id, client_secret):
 
 
 @pytest.mark.mais_tests
```
edsu, line 78: We can remove these markers now since we don't need to skip them. Also the we don't need this anymore:

https://github.com/sul-dlss/rialto-airflow/blob/ab179f369dae5877fe497cadb329b5f3c2ee4575/pyproject.toml#L31

edsu (approved):
I left some suggestions, but nothing critical.

test/test_mais.py
peetucket: removed

rialto_airflow/mais.py
peetucket: adjusted -- will also go do the same thing the mais_orcid_client gem ... i think they should work the same way so if one breaks, they all break together

PR #384, opened 2025-05-08
add pubmed harvest step and add it to the dag

### LINKED ISSUE #166: Write harvest_pub_med task
The `harvest_pub_med` will need to:

- Query the Pubmed API by ORCID iD

peetucket:
Fixes #166 - harvest publications from pubmed


compose.yaml
```diff
@@ -93,6 +93,7 @@ x-airflow-common:
     AIRFLOW_VAR_SUL_PUB_HOST: ${AIRFLOW_VAR_SUL_PUB_HOST}
     AIRFLOW_VAR_SUL_PUB_KEY: ${AIRFLOW_VAR_SUL_PUB_KEY}
     AIRFLOW_VAR_WOS_KEY: ${AIRFLOW_VAR_WOS_KEY}
+    AIRFLOW_VAR_PUBMED_KEY: ${AIRFLOW_VAR_PUBMED_KEY}
```
peetucket, line 96: key is already in vault and set by puppet

rialto_airflow/dags/harvest.py
```diff
@@ -226,6 +243,8 @@ def upload_data_quality_files(snapshot):
 
     snapshot = load_authors(snapshot)
 
+    pubmed_harvest(snapshot)
```
edsu, line 246: This is going to change once #387 is merged. I think as this stands since the return value of calling `pubmed_harvest()` isn't assigned to a variable that is a dependency for the later steps in the workflow those later steps could potentially run before this has completed?

rialto_airflow/harvest/pubmed.py
```diff
@@ -0,0 +1,196 @@
...
+    """
+    if not pmids or not isinstance(pmids, list):
+        return None
+
+    query = "id=" + "&id=".join(pmids)
+
+    http = requests.Session()
+
+    full_url = (
+        f"{BASE_URL}{FETCH_PATH}&api_key={os.environ.get('AIRFLOW_VAR_PUBMED_KEY')}"
```
edsu, line 115: Maybe the key could be looked up at the top of the file where you are setting some constants instead of getting from the environment over and over?

rialto_airflow/harvest/pubmed.py
```diff
@@ -0,0 +1,196 @@
...
+def publications_from_pmids(pmids):
+    """
+    Returns full pubmed records given a list of PMIDs.
+    """
+    if not pmids or not isinstance(pmids, list):
+        return None
+
+    query = "id=" + "&id=".join(pmids)
+
+    http = requests.Session()
```
edsu, line 112: Establishing a session can be useful for not setting up and tearing down the socket for every request. But since you are doing it here in this function it will get garbage collected almost immediately. You might want to just call `requests.post` below on line 118?

rialto_airflow/harvest/pubmed.py
```diff
@@ -0,0 +1,196 @@
...
+
+    params: Params = {"term": query}
+
+    http = requests.Session()
+
+    full_url = (
+        f"{BASE_URL}{SEARCH_PATH}&api_key={os.environ.get('AIRFLOW_VAR_PUBMED_KEY')}"
+    )
+    logging.info(f"searching pubmed with {params}")
+    resp: requests.Response = http.get(full_url, params=params, headers=HEADERS)
```
edsu, line 144: it might be worth calling `resp.raise_for_status()` here which will make sure an exception is thrown unless we get back a HTTP Status that went ok: 

https://3.python-requests.org/api/#requests.Response.raise_for_status

rialto_airflow/harvest/pubmed.py
```diff
@@ -0,0 +1,196 @@
...
+                            .values(publication_id=pub_id, author_id=author.id)
+                            .on_conflict_do_nothing()
+                        )
+
+                        jsonl_output.write(json.dumps(pubmed_pub) + "\n")
+
+    return jsonl_file
+
+
+def pmids_from_orcid(orcid):
```
edsu, line 92: We haven't been consistent about this, but to help with type-checking it might be good to add type annotations for the parameters. So like:

```python
def pmids_from_orcid(orcid: str) -> list[str]:
    ...
```

If you decide to do this, it would be appropriate for other functions in this file as well.

edsu (changes requested):
I noticed that the test coverage went down by quite a bit. I wonder if there are any obvious ways of increasing the coverage?

rialto_airflow/harvest/pubmed.py
peetucket: will do this

rialto_airflow/harvest/pubmed.py
peetucket: created a method to fetch the key

peetucket:
rebased from task groups PR and tested in localhost again

edsu (approved):
Looks good!

PR #387, opened 2025-05-10
Use task groups

### LINKED ISSUE #213: Use Airflow Task Groups
If we group our related DAG tasks with [Task Groups](https://www.astronomer.io/docs/learn/task-groups/) they should display better in the Airflow Graph diagram. This would group together:

- Harvesting by ORCID
- Harvesting by DOI
- Publishing Dashboards

edsu:
Group the Airflow tasks into task groups to make the workflow graph display a bit easier to understand. This also simplifies how the workflow dependencies are expressed in code, which seems like a big improvement.

Currently the workflow graph looks like:

Screenshot 2025-05-10 at 7 37 31 AM

WIth task groups it looks like the below screenshot. All the task steps are dependent on the initial setup task which creates the snapshot directory and database; so the graph apparently can't be expressed as a simple linear line.

Screenshot 2025-05-10 at 7 27 01 AM

You can expand different task groups (the blue boxes) to see the tasks they contain:

Screenshot 2025-05-10 at 7 27 12 AM

Screenshot 2025-05-10 at 7 27 18 AM

Closes #213

### DIFF
--- a/rialto_airflow/dags/harvest.py
+++ b/rialto_airflow/dags/harvest.py
@@ -3,8 +3,9 @@ import logging
 from pathlib import Path
 import shutil
 
-from airflow.decorators import dag, task
+from airflow.decorators import dag, task, task_group
 from airflow.models import Variable
+from airflow.models.xcom_arg import XComArg
 from honeybadger import honeybadger  # type: ignore
 
 from rialto_airflow import funders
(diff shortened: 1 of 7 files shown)


rialto_airflow/dags/harvest.py
```diff
@@ -220,50 +221,22 @@ def upload_data_quality_files(snapshot):
...
-    )
-
-    # link up the renaming tasks running the ones in tuples in parallel
-
-    (
-        (distilled_pubs, linked_pubs)
-        >> publish_open_access(snapshot)
-        >> upload_open_access_files(snapshot)
-    )
+    snapshot = setup()
```
edsu, line 232: Using the task groups meant that the linking together of dependencies could be simplified.

rialto_airflow/harvest/dimensions.py
```diff
@@ -211,4 +212,4 @@ def fill_in(snapshot: Snapshot, jsonl_file: Path) -> Path:
 
     logging.info(f"filled in {count} publications")
```
edsu, line 214: It's more important that we return the path to the new JSON file than it is the snapshot that was passed in?

rialto_airflow/dags/harvest.py
```diff
@@ -83,92 +84,87 @@ def load_authors(snapshot):
...
     @task()
-    def sul_pub_harvest(snapshot):
+    def sulpub_harvest(snapshot):
         """
         Harvest data from SUL-Pub.
         """
-        jsonl_file = sul_pub.harvest(
-            snapshot, sul_pub_host, sul_pub_key, limit=harvest_limit
-        )
+        sul_pub.harvest(snapshot, sul_pub_host, sul_pub_key, limit=harvest_limit)
```
edsu, line 115: This is a new task group for all the harvest tasks.

rialto_airflow/dags/harvest.py
```diff
@@ -83,92 +84,87 @@ def load_authors(snapshot):
...
 
-        return jsonl_file
+    @task_group()
+    def harvest_pubs(snapshot):
+        dimensions_harvest(snapshot)
+        openalex_harvest(snapshot)
+        wos_harvest(snapshot)
+        sulpub_harvest(snapshot)
 
     @task()
```
edsu, line 123: We don't actually need all the jsonl files. We were previously using them to express dependencies between tasks, but that is now no longer needed since we can use the task group as a workflow dependency.

rialto_airflow/dags/harvest.py
```diff
@@ -220,50 +221,22 @@ def upload_data_quality_files(snapshot):
...
-    snapshot = load_authors(snapshot)
-
-    sul_pub_jsonl = sul_pub_harvest(snapshot)
+    @task_group()
+    def upload(snapshot):
+        upload_open_access_files(snapshot)
+        upload_data_quality_files(snapshot)
 
-    dimensions_jsonl = dimensions_harvest(snapshot)
+    # link up dag tasks and task groups
```
edsu, line 230: All of this gets greatly simplified since we are linking the tasks together with `>>` below instead of using the task outputs to express workflow dependencies.

PR #389, opened 2025-05-12
add pubmed fill_in task and tests

### LINKED ISSUE #202: Create "fill-in" task for PubMed
Similar to #194 we want to query PubMed by DOI for any publications that lack a `pubmed_json` value, once the initial harvesting by ORCID has finished.

See #166 for harvesting from pub-med -- wait until that is done

peetucket:
Fixes #202, add Pubmed fill in task and tests.

Also, add a custom codecov.yml file

~~HOLD: need to verify the pubmed search by DOI works as expected.~~

~~When running on my laptop, the search for 50 DOIs yielded way too many PMIDs, suggesting it didn't work as expected.~~  Searching for DOIs one by one seems to work better.

### DIFF
--- a/rialto_airflow/harvest/pubmed.py
+++ b/rialto_airflow/harvest/pubmed.py
@@ -3,6 +3,7 @@ import logging
 import os
 import re
 from pathlib import Path
+import time
 
 import requests
 import xmltodict
@@ -153,9 +154,17 @@ def pmids_from_dois(dois: list[str]) -> list[str]:
     Returns PMIDs associated with given list of DOIs.
     """
 
-    query = " OR ".join([f'"{s}"' for s in dois])
+    # Note that we need to do these one by one, since Pubmed doesn't support searching for multiple DOIs
+    pmids = []
+    for doi in dois:
+        # look up the PMID given a DOI
+        pmid_results = _pubmed_search_api(f"{doi}")
+        # assuming we get one result, this is probably the PMID we want
+        if len(pmid_results) == 1:
+            pmids.append(pmid_results[0])
+        time.sleep(0.5)  # add a small delay to avoid hitting the API too hard
 
-    return _pubmed_search_api(f"{query}")
+    return pmids
 
 
 def publications_from_pmids(pmids: list[str]) -> list[str]:
--- a/rialto_airflow/harvest/pubmed.py
+++ b/rialto_airflow/harvest/pubmed.py
@@ -8,8 +8,7 @@ import requests
 import xmltodict
 
 from typing import Optional, Dict, Union
-
-# from sqlalchemy import select, update
+from sqlalchemy import select, update
 from sqlalchemy.dialects.postgresql import insert
 
 from rialto_airflow.database import (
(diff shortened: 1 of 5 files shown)


rialto_airflow/harvest/pubmed.py
```diff
@@ -93,7 +92,52 @@ def harvest(snapshot: Snapshot, limit=None) -> Path:
     return jsonl_file
 
 
-def pmids_from_orcid(orcid):
```
peetucket, line 96: this method didn't go away, i just added a type check for the return value, but the diff moved this change down below the new `fill_in` method

rialto_airflow/harvest/pubmed.py
```diff
@@ -104,6 +148,16 @@ def pmids_from_orcid(orcid):
...
 
 
+def pmids_from_dois(dois: list[str]) -> list[str]:
+    """
+    Returns PMIDs associated with given list of DOIs.
+    """
+
+    query = " OR ".join([f'"{s}"' for s in dois])
+
+    return _pubmed_search_api(f"{query}")
```
peetucket, line 158: query Pubmed API for PMIDs given DOIs (it's basically a term search quoting the DOIs and ORing them together

test/harvest/test_pubmed.py
```diff
@@ -83,6 +203,36 @@ def test_pubmed_search_no_publications():
...
+
+
+def test_pubmed_search_dois_no_publications():
+    """
+    This is a live test of the Pubmed Search API to ensure no results are returned for DOIs with no publications.
+    """
+    # No publications should be found for these dois
+    dois = ["bogus-doi-1", "bogus-doi-2"]
+    pmids = pubmed.pmids_from_dois(dois)
+    assert len(pmids) == 0, "found no publications"
```
peetucket, line 233: live tests for the new Pubmed API call that queries by DOI

test/harvest/test_pubmed.py
```diff
@@ -181,6 +331,101 @@ def test_harvest_when_doi_exists(
         assert pub.authors[1].orcid == "https://orcid.org/0000-0000-0000-0002"
 
 
+def test_fill_in(snapshot, test_session, mock_publication, caplog, monkeypatch):
+    caplog.set_level(logging.INFO)
```
edsu, line 335: Do we need this? Maybe it's left over from some testing?

test/harvest/test_pubmed.py
```diff
@@ -44,11 +47,85 @@ def existing_publication(test_session):
         return pub
 
 
+def mock_jsonl(path):
```
edsu, line 50: Does this actually need to be so detailed?

rialto_airflow/harvest/pubmed.py
```diff
@@ -104,6 +149,24 @@ def pmids_from_orcid(orcid):
...
+def pmids_from_dois(dois: list[str]) -> list[str]:
+    """
+    Returns PMIDs associated with given list of DOIs.
+    """
+
+    # Note that we need to do these one by one, since Pubmed doesn't support searching for multiple DOIs
+    pmids = []
+    for doi in dois:
+        # look up the PMID given a DOI
+        pmid_results = _pubmed_search_api(f"{doi}")
```
edsu, line 161: If you aren't formatting the doi you can just pass it in as a string?

```python
_pubmed_search_api(doi)
```

rialto_airflow/harvest/pubmed.py
```diff
@@ -104,6 +149,24 @@ def pmids_from_orcid(orcid):
...
+        # assuming we get one result, this is probably the PMID we want
+        if len(pmid_results) == 1:
+            pmids.append(pmid_results[0])
+        time.sleep(0.5)  # add a small delay to avoid hitting the API too hard
```
edsu, line 165: It'll be interesting to see how long this fill in step runs in prod...

edsu (approved):
This looks good. I left a few suggestions and questions. But I don't think they are critical to resolve before merging.

rialto_airflow/harvest/pubmed.py
peetucket: yes good catch, this was leftover from when i was actually using formatting to do something more complicated, but no longer needed

test/harvest/test_pubmed.py
peetucket: maybe?  this was in the other fill in tests, my assumption was it is there to ensure the log file level is set in a way that the assertions below would definitely work (i.e. if the default log level is changed elsewhere at some point, the assertions might start failing?)

test/harvest/test_pubmed.py
peetucket: i removed some data but wanted it to be at least somewhat similar to the json structure of an actual record

Issue #402, opened 2025-05-20
Dependencies and capistrano deploy


edsu:
It appears that the Docker image is not getting rebuilt when doing a `cap deploy` and there is a change in the `uv.lock` and `pyproject.toml`. 

I did a `bundle exec cap prod deploy` and saw that the harvest dag failed to load because of a missing dependency:

![Image](https://github.com/user-attachments/assets/8f78e191-47b8-4388-ba99-6a0e83ce7433)

I ssh'd to the server and could see the `uv.lock` contained the missing dependency:

```
rialto@sul-rialto-airflow-prod:~/rialto-airflow/current$ grep xmltodict uv.lock
    { name = "types-xmltodict" },
    { name = "xmltodict" },
    { name = "types-xmltodict", specifier = ">=0.14.0.20241009" },
    { name = "xmltodict", specifier = ">=0.14.2" },
name = "types-xmltodict"
sdist = { url = "https://files.pythonhosted.org/packages/05/28/f1c86d869d74342553c74bd88d6f615427b544fd06f30fdaa68cb4369cd1/types-xmltodict-0.14.0.20241009.tar.gz", hash = "sha256:9224c2422c5b6359cf826685b4ee50b14dc2cb9134561ab793ef6b03dd7108e1", size = 3224, upload-time = "2024-10-09T02:43:18.364Z" }
    { url = "https://files.pythonhosted.org/packages/15/5e/6d81048a06f956ccec74c61fe5c30784e6a865876dc4ab02e99099e4ee0c/types_xmltodict-0.14.0.20241009-py3-none-any.whl", hash = "sha256:92812e17ffa9171416b35806cb5f4ed3f8f52b6724b2c555e4733e902ef4afd0", size = 2859, upload-time = "2024-10-09T02:43:16.966Z" },
name = "xmltodict"
sdist = { url = "https://files.pythonhosted.org/packages/50/05/51dcca9a9bf5e1bce52582683ce50980bcadbc4fa5143b9f2b19ab99958f/xmltodict-0.14.2.tar.gz", hash = "sha256:201e7c28bb210e374999d1dde6382923ab0ed1a8a5faeece48ab525b7810a553", size = 51942, upload-time = "2024-10-16T06:10:29.683Z" }
[...]

PR #406, opened 2025-05-21
see if we can get docker to build on deploy

peetucket:
Maybe deals with #402 ? 

 1. Looked at https://github.com/sul-dlss/dlss-capistrano-docker
 2. Tried changing this setting
 3. Deployed the app and saw it building the docker container before bringing it up
 4. Tried adding a new dependency and importing it in harvest dag and deploying
 5. All worked, so maybe ok?


edsu (approved):
🤞

PR #408, opened 2025-05-22
Validate tableau 381, closes #381

### LINKED ISSUE #381: Add validate_calculation task for ORCID dashboard
Automatically testing the Tableau dashboard against the published csv files is proving challenging (see notes [here](https://github.com/sul-dlss/rialto-airflow/issues/310)) so we have decided to adopt a hybrid approach for now. We want to add a report step similar to DLME that reads the published csv files, calculates some expected values, and sends the report to Rochelle's email address. The ORCID test should confirm the following:

- [x] the count for the date April, 2024 should be 806
- [x]  calculate the ORCID percent of the faculty in the bar chart
- [ ] calculate the no ORCID percent of the MS students in the bar chart
- [ ] calculate the percent for faculty in the Engineering school in the highlight table
- [ ] calculate percent of PhD students in SLAC (there should be no PhD students in SLAC)
- [ ] similar calculates by department
- [ ] get the third name from the ORCID adoption individual table

jacobthill:
Adds the validate task to the Publish ORCID DAG.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -26,6 +26,7 @@ dependencies = [
     "sqlalchemy>=1.4.36,<2.0", # airflow still incompatible with 2.x, see https://github.com/apache/airflow/issues/28723
     "xmltodict>=0.14.2",
     "types-xmltodict>=0.14.0.20241009",
+    "dominate>=2.9.1",
 ]
 
 [tool.pytest.ini_options]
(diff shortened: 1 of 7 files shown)


peetucket (dismissed):
I'm going to take a stab at reverting some of the changes this PR would make so they match what is on main and then push up a new commit

PR #409, opened 2025-05-22
Extract publish steps into DAGs

### LINKED ISSUE #317: Separate Publish from Harvest DAG
Since the Harvest dag can take days to run to completion it would be helpful to be able to run the publish DAG separately so we can adjust how it is assembling and publishing CSV data for various dashboards. It may make sense to have different DAGs for different reports if we want to run them separate from each other?

We will need some utility to determine the latest snapshot to use for reporting. This might involve the Harvest DAG writing some sort of completion file to the snapshot directory.

edsu:
This commit extracts the publish steps from the harvest DAG into separate DAGs that can be run independently.

At the moment it is scheduled to run weekly. As a separate task we could consider having the publish steps get run automatically once a harvest is finished.

Closes #317


rialto_airflow/dags/harvest.py
```diff
@@ -192,66 +189,8 @@ def post_process(snapshot):
         link_funders(snapshot)
```
edsu, line 189: This is the publish functionality that was extracted into separate DAGs.

rialto_airflow/dags/publish_data_quality.py
edsu, line 1: This is the new DAG for publishing the data quality dashboard files.

rialto_airflow/dags/publish_open_access.py
edsu, line 1: This is the new DAG for publishing openaccess dashboard files.

rialto_airflow/snapshot.py
edsu, line 1: The Snapshot class was heavily refactored to allow you to create a snapshot independent of the constructor, since I wanted to be able to create a new Snapshot object for an already existing snapshot directory. I also wanted to be able to pick out the latest completed snapshot.

rialto_airflow/snapshot.py
```diff
@@ -11,16 +13,81 @@ class Snapshot:
...
+                continue
+
+            # TODO: store the database name in snapshot.json?
+            database_name = f"rialto_{snapshot_path.name}"
+
+            return Snapshot(snapshot_path, snapshot_name, database_name)
+
+        # no complete snapshot was found
+        return None
+
```
edsu, line 64: This method marks a Snapshot as complete. It is called at the end of the harvest DAG.

test/conftest.py
```diff
@@ -120,4 +120,4 @@ def mock_association(test_session, mock_publication, mock_authors):
```
edsu, line 120: Many of the tests changed slightly since the Snapshot constructor changed.

test/test_database.py
```diff
@@ -9,11 +9,6 @@
 from rialto_airflow.snapshot import Snapshot
```
edsu, line 10: We can use the snapshot fixture from conftest.py here.

rialto_airflow/snapshot.py
```diff
@@ -11,16 +13,81 @@ class Snapshot:
...
+        data_dir = data_dir
+
+        snapshots_dir = data_dir / "snapshots"
+        if not snapshots_dir.is_dir():
+            return None
+
+        snapshot_names = [p.name for p in snapshots_dir.iterdir()]
+
+        for snapshot_name in reversed(sorted(snapshot_names)):
+            # ignore things that might be here that aren't the right format
```
peetucket, line 48: I guess sorting by folder name is sufficient to get the late time stamped one

edsu: I wanted to get them in reverse order, so the oldest was picked first. I can add a comment about this if it doesn't make sense?

peetucket: I understood the reverse part, was more of a question about if the way Python does sorting of arrays works the way we'd expect with the way snapshot folders are named ... at first thought I guess it does (e.g. `20250512193014` etc. always has bigger names being later in time?)

edsu: Yes, I think that's how it should work. The initial list on line 45 could be in any order. Then the list is sorted, and then reversed, so we'd get them in descending order: 

```python
['20250512193014', '20250511123033', '20250411082119', ...]
```

peetucket (approved):

PR #413, opened 2025-05-22
Extract publish steps into DAGs

### LINKED ISSUE #317: Separate Publish from Harvest DAG

peetucket:
This commit extracts the publish steps from the harvest DAG into separate DAGs that can be run independently.

At the moment it is scheduled to run weekly. As a separate task we could consider having the publish steps get run automatically once a harvest is finished.

Closes #317

Same as #409 but rebased from main


edsu:
This is no longer needed I don't think.

PR #428, opened 2025-06-02
Join Funders table

### LINKED ISSUE #423: Disk space fills up in publication publish steps
`sqlalchemy.exc.OperationalError: (psycopg2.errors.DiskFull) could not write to file "base/pgsql_tmp/pgsql_tmp120947.396": No space left on device`

see https://stanfordlib.slack.com/archives/C0RK5EM9N/p1748888615249969

When running the publish_publications DAG, the disk space fills up.  Speculation: the group by clause in the publication publish steps is making for a very big temporary query, even though execution options sets to only return 100 rows, it maybe has to scan the whole table

This is happening after this log message:

```
[2025-06-02, 17:53:34 UTC] {publication.py:226} INFO - starting to write contributions by school/department /opt/airflow/data/snapshots/20250525000004/publication-dashboard/contributions-by-school-department.csv
```

edsu:
The school and department queries report on data from the Funders table without joining on it, which results in a cartesian product that fills up the database sort space. This commit adds the join, and adjusts the tests to sort by the school and department as appropriate to ensure that the rows can be reliably tested (they are not dependent on the implicit sort order of the query results).

Fixes #423 🤞

This PR also includes several changes to publication_utils.py to improve the reliability of working with the extracted JSON, as well as a performance improvement to the distiller, which now caches JSON Paths that have been parsed.

### DIFF
--- a/rialto_airflow/publish/publication.py
+++ b/rialto_airflow/publish/publication.py
@@ -65,7 +65,7 @@ def write_publications(snapshot) -> Path:
                 )
                 .join(Author, Publication.authors)  # type: ignore
                 .group_by(Publication.id)
-                .execution_options(yield_per=100)
+                .execution_options(yield_per=10_000)
             )
 
             for row in session.execute(stmt):
@@ -81,6 +81,7 @@ def write_publications(snapshot) -> Path:
                         "faculty_authored": "faculty" in row.primary_role,
                     }
                 )
+                output.flush()
 
         logging.info(f"finished writing publications {csv_path}")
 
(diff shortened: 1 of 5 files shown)


peetucket (approved):
Approve (hoping it works as expected!)

rialto_airflow/publish/publication.py
edsu, line 1: We need to join on Funders or else 💥

rialto_airflow/publish/publication_utils.py
```diff
@@ -26,7 +26,8 @@ def _openalex_journal(openalex_json):
     try:
         primary_location = openalex_json["primary_location"]
```
edsu, line 27: sometimes primary_location in OpenAlex data is None -- so we need to guard against that.

test/publish/test_publication.py
```diff
@@ -178,7 +178,7 @@ def test_write_contributions_by_school(test_session, snapshot, dataset, caplog):
     assert len(df) == 3
 
     # sort it so we know what is in each row
-    df = df.sort_values(["doi"])
```
edsu, line 181: ordering by the primary school too means we can expect the results in a particular order. adding the join to Funders caused the order to fall out of alignment with the testing.

test/publish/test_publication_utils.py
```diff
@@ -178,6 +178,20 @@ def test_journal_with_pubmed_pmid():
     assert pub_utils._pmid(row) == "36857419"
```
edsu, line 180: This test is here to appease test coverage...

edsu:
I'm choosing to ignore the warning about test coverage here. It seems related to the exception handling being broadened to include TypeError, which gets thrown when you treat a str like a dictionary in some cases during JSON extraction. We are already testing the

Issue #400, opened 2025-05-20
Investigate Pubmed API to see if we can query by DOI more than one at a time


peetucket:
Current fill-in task for Pubmed takes a very long time since we currently search by DOI one at a time.  Investigate just to be sure there aren't other techniques we can use to query for multiple DOIs at once and discover PMIDs.

PR #436, opened 2025-06-09
disable pubmed fillin

peetucket:
It is running very slowly and we need to return to address it later (see #400)


rialto_airflow/dags/harvest.py
```diff
@@ -147,7 +147,12 @@ def fill_in_pubmed(snapshot):
         """
         Fill in Pubmed data for DOIs from other publication sources.
         """
-        pubmed.fill_in(snapshot)
```
edsu, line 150: I think commenting this out here will make it look like the task is still running? How about instead we comment out the call to `fill_in_pubmed(snapshot)` in `fill_in`. Then it will just not be present in the workflow?

edsu (changes requested):

rialto_airflow/dags/harvest.py
peetucket: changed

PR #448, opened 2025-06-24
add sul-pub-counts.csv file to data quality dashboard

### LINKED ISSUE #443: Generate sul-pub-counts.csv with unique DOI counts exclusive to SUL-Pub
## Description
To preserve visibility into SUL-Pub’s unique contributions, we need a separate file showing the count of DOIs found exclusively in SUL-Pub and nowhere else.

## Request
- Implement a process to identify DOIs that appear **only in SUL-Pub** and not in any other source.  
- Generate a new CSV file named `sul-pub-counts.csv` containing a single row with the count of these unique DOIs.  
- This file will provide a clear, separate metric of SUL-Pub’s unique value without mixing it into `source-counts.csv`.

peetucket:
Fixes #443

### DIFF
--- a/test/publish/test_data_quality.py
+++ b/test/publish/test_data_quality.py
@@ -417,13 +417,43 @@ def test_write_total_source_count(test_session, snapshot):
 
         rows = df.to_dict("records")
         assert (rows[0]["source"]) == "Dimensions"
-        assert 220 <= rows[0]["total_count"] <= 280, "expect ~25% for Dimensions"
+        assert 210 <= rows[0]["total_count"] <= 290, "expect ~25% for Dimensions"
         assert (rows[1]["source"]) == "Openalex"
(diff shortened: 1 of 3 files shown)


test/publish/test_data_quality.py
```diff
@@ -426,6 +426,36 @@ def test_write_total_source_count(test_session, snapshot):
...
+def test_write_sulpub_source_count(test_session, snapshot):
+    # create some random data
+    with test_session.begin() as session:
+        for i in range(0, 1000):
+            session.add(
+                Publication(
+                    doi=f"10.000/00000{i}",
+                    pub_year=2024,
+                    sulpub_json={"a": "b"}
+                    if randint(0, 3) == 1
```
edsu, line 438: Instead of doing it randomly how about every 4th  publication gets wos_json, then our test can be precise.

```python
if i % 4 == 0

peetucket:
Adjusted query and tests to ensure it works as described

Issue #451, opened 2025-07-07
COMET metadata augmentation examples


edsu:
We're looking for ~10 records to demonstrate RIALTO metadata augmentation for the [COMET project](https://www.cometadata.org/). 

Ideally looking for:
*  Records where the Stanford distilled record is richer than the “naked” record at the DOI registrar (CrossRef or DataCite). Funder and APC could be a natural augmentation since we are currently only source for this.
*  Records where we can add augmented affiliation data: Stanford / SLAC / Stanford Medicine. Adding ROR would be icing on the cake.

edsu:
This was done in this [Jupyter notebook](https://github.com/sul-dlss-labs/rialto-notebooks/blob/main/comet.ipynb).

PR #452, opened 2025-07-11
Collect data from Crossref

edsu:
This PR adds fetching metadata from the [Crossref API](https://api.crossref.org/swagger-ui/index.html#/limits) by DOI to the "fill in" task in the harvest DAG. Collecting the Crossref metadata was useful for exploratory work to support our participation in COMET #451 and I thought we might as well add it since the metadata does seem to have some value?

We can look up 40 DOIs with each API request, and there is a sleep of one second in between requests. Maybe the sleep is not needed since [they say](https://api.crossref.org/swagger-ui/index.html#limits) that they allow up to 50 requests per second, but I thought we'd just play it safe.

I ran this in production last weekend to work out whatever kinks there might be with DOI formatting and the API responses. For about 500K DOIs it took about 1.5 days to run to completion. I analyzed the log and the responses take 8 seconds on average to come back (std dev 6.15).

While we are doing full, non-incremental harvesting we may want to dial them back to running biweekly or monthly instead of every week?

### DIFF
--- a/rialto_airflow/harvest/crossref.py
+++ b/rialto_airflow/harvest/crossref.py
@@ -37,9 +37,6 @@ def fill_in(snapshot: Snapshot) -> Path:
                 dois = [normalize_doi(row.doi) for row in rows]
                 for crossref_pub in get_dois(dois):
                     doi = normalize_doi(crossref_pub.get("DOI"))
-                    if doi is None:
-                        logging.warning("unable to determine what DOI to update")
-                        continue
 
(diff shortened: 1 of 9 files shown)


rialto_airflow/harvest/crossref.py
edsu, line 1: This is the new module for collecting data from the Crossref API. There's no `harvest()` method since we aren't looking up publications by ORCID. We simply walk through all the DOIs that were collected and fetch their metadata, 40 at a time.

peetucket (approved):

PR #458, opened 2025-07-28
Normalize spaces in DOIs and prevent lookup

edsu:
We had a situation where the DOI "doi:10.1016/ j.xkme.2022.100435" was being looked up. Our normalize_doi() function was stripping leading and trailing whitespace but not whitespace inside the string. This change adds the stripping of whitespace, and also ensures that we don't try to look them up in the Crossref API which throws a HTTP 400 error that stops the workflow.

### DIFF
--- a/rialto_airflow/harvest/crossref.py
+++ b/rialto_airflow/harvest/crossref.py
@@ -67,7 +67,8 @@ def get_dois(dois: Iterable[str], tries=5) -> Iterable[dict]:
 
             # According to Crossref API error messages the DOI must be of the form:
             # doi:10.prefix/suffix where prefix is 4 or more digits and suffix is a string
-            if m := re.match(r"^doi:10\.(\d+)/.+$", doi):
+            # the string must not contain whitespace
+            if m := re.match(r"^doi:10\.(\d+)/\S+$", doi):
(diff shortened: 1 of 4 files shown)


rialto_airflow/harvest/crossref.py
```diff
@@ -67,7 +67,8 @@ def get_dois(dois: Iterable[str], tries=5) -> Iterable[dict]:
 
             # According to Crossref API error messages the DOI must be of the form:
             # doi:10.prefix/suffix where prefix is 4 or more digits and suffix is a string
-            if m := re.match(r"^doi:10\.(\d+)/.+$", doi):
+            # the string must not contain whitespace
+            if m := re.match(r"^doi:10\.(\d+)/\S+$", doi):
```
peetucket, line 71: is this just an additional safeguard, because in theory, the cleanup method below should prevent whitespace from ever being sent?

edsu: Yeah, that's true. It was here so I could move an existing DAG run along last week. It doesn't seem like it will do any harm to have it?

peetucket: yup, no harm, just making sure i understood correctly!

PR #460, opened 2025-07-28
Batch query Pubmed by DOI instead of one by one

### LINKED ISSUE #400: Investigate Pubmed API to see if we can query by DOI more than one at a time

peetucket:
Fixes #400 - batch query the pubmed by DOI instead of one by one

Seems to be much faster in localhost

### DIFF
--- a/rialto_airflow/harvest/pubmed.py
+++ b/rialto_airflow/harvest/pubmed.py
@@ -102,7 +102,7 @@ def fill_in(snapshot: Snapshot):
                 select(Publication.doi)  # type: ignore
                 .where(Publication.doi.is_not(None))  # type: ignore
                 .where(Publication.pubmed_json.is_(None))
-                .execution_options(yield_per=50)
+                .execution_options(yield_per=75)
             )
 
             for rows in select_session.execute(stmt).partitions():
--- a/rialto_airflow/harvest/pubmed.py
+++ b/rialto_airflow/harvest/pubmed.py
@@ -3,7 +3,6 @@ import logging
 import os
 import re
 from pathlib import Path
-import time
 
 import requests
 import xmltodict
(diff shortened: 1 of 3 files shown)


rialto_airflow/harvest/pubmed.py
```diff
@@ -3,7 +3,6 @@
 import os
 import re
 from pathlib import Path
-import time
```
peetucket, line 6: test removing the sleep between requests

edsu (approved):
Did you do any testing to see how many can be looked up at once? If you can do more than 50 you could consider changing the select to returning more DOIs in each batch?

peetucket:
> Did you do any testing to see how many can be looked up at once? If you can do more than 50 you could consider changing the select to returning more DOIs in each batch?

good question - i didn't .... let me do some more testing locally

it's possible this may be limited by the size of the querystring we can send to the API

edsu (approved):
It seems to work fine with a batch size of 75 DOIs too. I tested with  HARVEST_LIMIT=10000.

peetucket:
> It seems to work fine with a batch size of 75 DOIs too. I tested with HARVEST_LIMIT=10000. I'm fine with either way. This will be nice to have for the data quality reporting!

I'm guessing the limit is the size of the query string passed to the API.  Somewhere between 50 and 100 we run into the limit depending on how big the DOI strings are.  I'll bump to 75 but we would want to keep an eye on the logs the first couple of times as if the limit is exceeded, it will log the error and move on (so it may get missed).

PR #464, opened 2025-07-30
Build a static website

edsu:
This commit adds a publish_website DAG that will generate a static website using the https://github.com/sul-dlss-labs/rialto-site repository and publish it to the website subdirectory of \$DATA_DIR

When running in dev the static site is served up using nginx (running in Docker) at http://localhost:80

TBD: where to make the website available from


Dockerfile
```diff
@@ -1,7 +1,14 @@
 FROM apache/airflow:2.10.4-python3.12
 
 USER root
```
edsu, line 3: Install node and yarn into the Docker container, so that we can use them to build the static site.

compose.yaml
```diff
@@ -319,5 +319,13 @@ services:
       airflow-init:
         condition: service_completed_successfully
```
edsu, line 321: In development we can serve up the static site using nginx. If we had another webserver that was running that could serve the website out of the Rialto volume that would work for our QA, Stage and Production environments?

rialto_airflow/dags/website.py
edsu, line 1: This module runs git and yarn to build the website from https://github.com/sul-dlss-labs/rialto-site.

test/nginx.conf
edsu, line 1: This nginx config is only used in development, to make the static site available on port 80.

test/test_website.py
edsu, line 1: We could imagine better tests here, if the database was loaded up with some fixture data that would generate .json files that we could check.

rialto_airflow/plugins/site.py
edsu, line 1: This is an Airflow plugin to make the static website available. We probably wouldn't want to do this long term and would publish the website on an Apache instance running elsewhere?

edsu:
I think we've decided to move in a different direction, using a Rails app that can control visibility of different visualizations, which would be difficult to do in a static website.

PR #490, opened 2025-08-20
[initial reporting database implementation] store publications.csv data as a table in a stably named DB

### LINKED ISSUE #456: Research how to store data for visualization in postgres
We would need to move logic for munging the data from Airflow into Tableau, or write new tables/views for Tableau to access it.

- consider how we would store data (instead of flat csv file, an equivalent flat postgres table)
- create a table like this, see if tableau visualization can read from it likes it reads from a csv

pre-req #455

jmartin-sul:
to test this locally, besides unit tests, i did a small `harvest` DAG run with this branch, followed by a `publish_publications` DAG run (i commented out the upload step in the `publish_publications` DAG so i wouldn't inadvertently step on the shared drive's CSVs).  i saw the new table get populated with what looked like the contents that'd go in `publications.csv`, and using tableau desktop locally, i was able to make a simple graph visualizing DOI count vs pub year.

~~i'm not set on any of the decomposition or code placement choices in this PR.  maybe i could show it quickly as a post-standup discussion topic, and we could decide what if anything we like and want to keep, what we want to change direction on, and how to move this forward?~~  after some discussion and some naming tweaks, we've decided to push this forward as the start of a pattern for storing reporting data in database tables instead of CSV files.

closes #456

### DIFF
--- a/rialto_airflow/publish/publication.py
+++ b/rialto_airflow/publish/publication.py
@@ -2,15 +2,50 @@ import logging
 from csv import DictWriter
 from pathlib import Path
 
-from sqlalchemy import select, func
+from sqlalchemy import select, func, Boolean, Column, Integer, String
+from sqlalchemy.dialects.postgresql import insert
+from sqlalchemy.orm import declarative_base  # type: ignore
+from sqlalchemy.types import DateTime
 
-from rialto_airflow.database import get_session, Publication, Author, Funder
+from rialto_airflow.database import (
+    RIALTO_REPORTS_DB_NAME,
+    create_schema,
+    get_session,
+    utcnow,
+    Publication,
+    Author,
+    Funder,
+)
 from rialto_airflow.utils import get_types, get_csv_path, normalize_pmid
 import rialto_airflow.publish.publication_utils as pub_utils
 
 
+ReportsSchemaBase = declarative_base()
+
+
+class PublicationAugmented(ReportsSchemaBase):  # type: ignore
+    __tablename__ = "publication_augmented"
+
+    id = Column(Integer, primary_key=True, autoincrement=True)
+    doi = Column(String, unique=True)
+    pub_year = Column(Integer)
+    apc = Column(Integer)
+    open_access = Column(String)
+    types = Column(String)
+    funders = Column(String)
+    federally_funded = Column(Boolean)
+    academic_council_authored = Column(Boolean)
+    faculty_authored = Column(Boolean)
+    created_at = Column(DateTime, server_default=utcnow())
+    updated_at = Column(DateTime, onupdate=utcnow())
+
+
 def google_drive_folder() -> str:
     return "publication-dashboard"
+
+
+def init_reports_data_schema() -> None:
+    create_schema(RIALTO_REPORTS_DB_NAME, ReportsSchemaBase)
 
 
 def write_publications(snapshot) -> Path:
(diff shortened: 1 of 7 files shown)


jmartin-sul:
one more thing for consideration: i recall that we were hesitant (maybe even blocked by some other dependency?) to add [Alembic](https://alembic.sqlalchemy.org/en/latest/).  but, if we're going to have at least one stably named database with a schema that's likely to evolve over time, we'll probably need to do schema and data migrations.  and one other possibly high priority enhancement (per standup discussion today) is incremental updates (instead of a full re-harvest to a new DB each time), which will also prod us towards needing migrations.

rialto_airflow/publish/publication.py
```diff
@@ -2,17 +2,52 @@
...
+    created_at = Column(DateTime, server_default=utcnow())
+    updated_at = Column(DateTime, onupdate=utcnow())
+
+
 def google_drive_folder() -> str:
     return "publication-dashboard"
 
 
+def init_distilled_data_schema() -> None:
+    create_schema(DISTILLED_DATA_DB_NAME, DistillSchemaBase)
```
jmartin-sul, line 48: i had initially been thinking we'd need to truncate the publication table, similar to how we write a fresh full copy of the CSV every time.  but maybe this is ok, because `doi` column has a unique constraint, and we have `on_conflict_do_nothing()` as the `insert` behavior?

jmartin-sul: https://www.postgresql.org/docs/current/sql-truncate.html

rialto_airflow/publish/publication.py
```diff
@@ -70,22 +105,28 @@ def write_publications(snapshot) -> Path:
...
+                }
+                csv_output.writerow(row_values)
                 output.flush()
 
+                with get_session(DISTILLED_DATA_DB_NAME).begin() as insert_session:
+                    insert_session.execute(
+                        insert(PublicationAugmented)
+                        .values(**row_values)
+                        .on_conflict_do_nothing()
+                    )
```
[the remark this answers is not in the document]
jmartin-sul: after changes from yesterday's discussion, this now rewrites the table on every publish run, so it will remove publications if need be

test/publish/test_publication.py
```diff
@@ -147,6 +185,7 @@ def test_dataset(test_session, dataset):
         assert len(pub.funders) == 2
 
 
+@pytest.mark.usefixtures("setup_teardown_distilled_schema")
 def test_write_publications(test_session, snapshot, dataset, caplog):
```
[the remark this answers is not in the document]
jmartin-sul: test expectations added

rialto_airflow/publish/publication.py
```diff
@@ -2,17 +2,52 @@
...
+    pub_year = Column(Integer)
+    apc = Column(Integer)
+    open_access = Column(String)
+    types = Column(String)
+    funders = Column(String)
+    federally_funded = Column(Boolean)
+    academic_council_authored = Column(Boolean)
+    faculty_authored = Column(Boolean)
+    created_at = Column(DateTime, server_default=utcnow())
+    updated_at = Column(DateTime, onupdate=utcnow())
```
edsu, line 40: I'm not sure we need `created_at` and `updated_at` for reporting, but it seems harmless (and maybe useful) to have them.

rialto_airflow/publish/publication.py
```diff
@@ -2,17 +2,52 @@
...
+    Funder,
+)
 from rialto_airflow.utils import get_types, get_csv_path, normalize_pmid
 import rialto_airflow.publish.publication_utils as pub_utils
 
 
+ReportsSchemaBase = declarative_base()
+
+
+class PublicationAugmented(ReportsSchemaBase):  # type: ignore
```
edsu, line 26: I wonder if we really need `Augmented` in the name here? I think this came up in discussion but we may find ourselves wanting to name these tables in such a way that they reflect what tableau visualization they are built for. But if we can get reuse of `PublicationAugmented` across visualizations that would be ideal!

edsu (approved):
I left another review comment ... not a blocker.

Issue #491, opened 2025-08-20
Test deduplication as a post-process step (spike)


lwrubel:
See #382 on deduplication for context. 

To identify and deduplicate publications that occur within a platform, use a strategy of querying by platform id (e.g. WOS ID). 

For pubs without DOIs, query on each pub's platform ID in its JSON (WOS ID, Dimensions ID, PMID, etc). If the count is > 1:
* get the authors and move the author relationship to the remaining publication. 
* delete the duplicate publications.

This step would happen before Funders are linked to Publications.

lwrubel:
Closed as spike was discussed in storytime and PR ready for review.

PR #495, opened 2025-08-21
Deduplication as a post-process step

lwrubel:
Sample code for deduplication as a post process step with only one platforms's identifiers (WOS). Relates to #491.

~I think there should be a better way to handle deleting the pub_author_association. The things I tried initially didn't work, so there's more to look at.~ I set up a delete cascade on the publication / author many to many relationship.

### DIFF
--- a/rialto_airflow/dags/harvest.py
+++ b/rialto_airflow/dags/harvest.py
@@ -171,7 +171,7 @@ def harvest():
         """
         Remove duplicates based on WOS UID.
         """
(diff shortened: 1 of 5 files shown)


edsu (approved):
As we add deduplication for other identifiers we might want to think about how some of this code could be abstracted a bit so we don't need to repeat the same logic over and over. But then again, it reads well, and maybe the logic will need to be slightly different in other cases. Time will tell...

Issue #492, opened 2025-08-20
Test deduplication during upsert (spike)


edsu:
See #382 on deduplication for context.

Currently we are letting the database deduplicate publications by DOI when we do the insert because:

1. We have a DOI column.
2. We have a unique constraint on the DOI column.

When we are doing inserts into the Publication table we use SQLAlchemy's "upsert" support to recover from a unique constraint violation and update the existing record rather than insert a new one with the same DOI.

Here's an example of inserting OpenAlex metadata into the Publication table, where the `publication_doi_key` constraint is caught.

https://github.com/sul-dlss/rialto-airflow/blob/cf4b6f6f44829077a145ab6573c09f65dd704db4/rialto_airflow/harvest/openalex.py#L58-L68

If we were to create unique indexes for other identifiers, either referencing them directly in the JSON, or as separate columns, could we catch these and do updates too?

So for example if we had a unique index on `wos_json -> 'id'` could we use that to prevent us from adding publications that lack a DOI, but have a shared Web of Science ID?

PR #497, opened 2025-08-22
Add constraints for WOS and Dimensions ID

edsu:
This commit illustrates how we can reduce duplication by adding unique constraints for platform specific identifiers like the Web of Science ID and Dimensions ID.

First it adds unique indexes:

- doi_wos_idx: (`DOI`, `wos_json ->> 'UID'`)
- doi_dim_idx: (`DOI`, `dim_json ->> 'id'`)

The DOI value is coalesced into an empty string, so that the index continues to work when a DOI is `NULL`.

Then when we are doing the "upsert" during harvesting we reference the relevant index, to catch unique constraint violations.

Unfortunately only one index can be used as a constraint during an upsert. Also, it only appears to be possible to reference an unique constraint by name, so the index name is looked up manually with get_index().

For all the tests to pass we will need similar new indexes for the upsert to reference when harvesting OpenAlex, Crossref, sul_pub. I had to relax the unique constraint on DOI, as it was getting thrown in the Dimensions and WoS testing here.

Refs #492

### DIFF
--- a/rialto_airflow/database.py
+++ b/rialto_airflow/database.py
@@ -2,7 +2,7 @@ import logging
 import os
 from functools import cache
 
-from sqlalchemy import Table, Boolean, Column, ForeignKey, Integer, String
+from sqlalchemy import Table, Boolean, Column, ForeignKey, Integer, String, Index
 from sqlalchemy import create_engine, text
 from sqlalchemy.dialects.postgresql import ARRAY, JSONB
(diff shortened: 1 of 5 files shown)


edsu:
We decided to table this approach in favor of having a more flexible deduplicate DAG task run after initial harvesting of publications is complete.

PR #512, opened 2025-08-28
Dedupe OpenAlex pubs

### LINKED ISSUE #503: Deduplicate by OpenAlex id where DOI is lacking
There are currently 65,649 rows in the publications table which have an `openalex_json` value but a null `doi`.  3,624 of those openalex ids have duplicate publications (8,427 publication rows total).

Add a step to deduplicate on the `"id"` key in the openalex_json following the approach in https://github.com/sul-dlss/rialto-airflow/pull/495. You may want to abstract and/or refactor as you add in this step.

lwrubel:
Resolves #503, deduplicating publications by OpenAlex id. Also does some light refactoring.


rialto_airflow/harvest/deduplicate.py
```diff
@@ -31,19 +44,63 @@ def remove_wos_duplicates(snapshot: Snapshot) -> int:
...
+        # Move author relationships to the first instance
+        for author in pub.authors:
+            session.execute(
+                insert(pub_author_association)
+                .values(publication_id=main_pub, author_id=author.id)
+                .on_conflict_do_nothing()
+            )
+
+        # Delete the duplicate
+        session.execute(delete(Publication).where(Publication.id == pub.id))  # type: ignore
```
peetucket, line 104: do/should these happen in a transaction?

lwrubel: These changes are part of the session that's passed into the function, and which is held open with the context manager `with get_session(snapshot.database_name).begin() as session:`

The context manager should handle committing or rolling back if there are any errors. See: https://docs.sqlalchemy.org/en/14/orm/session_basics.html#framing-out-a-begin-commit-rollback-block 

So I think this is covered? But possible there's something I don't understand with sqlalchemy.

peetucket: sounds good, my comment should not be taken as knowing something more than that !

lwrubel: I'm going to test this in stage all the same.

lwrubel:
@peetucket I made some changes after testing, mainly fixing what function the deduplicate task calls and clarifying some logging. Would you take a look again?

peetucket (approved):

PR #570, opened 2025-09-16
Fix the seed data loaded in orcid_integration_stats migration

### LINKED ISSUE #565: Load prod historic data into orcid_integration_stats
The `rialto_reports` migration that loads historic ORCID integration stats in the `orcid_integration_stats` table is only loading the Stage Google Sheet, which is incomplete. Instead, we need to load the prod historic data. 

Replace https://github.com/sul-dlss/rialto-airflow/blob/main/alembic/orcid-integration-stats-seed.csv with the prod data.

Write a migration that drops the current table data and loads what's in the seed CSV. You can use this migration as a template. https://github.com/sul-dlss/rialto-airflow/blob/main/alembic/versions/a7b78175dd8b_add_existing_stats_to_orcid_integration_.py

lwrubel:
Resolves #565 to get the correct full data seeded in the orcid_integration_stats table.


alembic/versions/4b7bdf4035f0_seed_full_historic_orcid_integration_.py
```diff
@@ -0,0 +1,53 @@
...
+            sa.column("read_only_scope", sa.Integer),
+            sa.column("read_write_scope", sa.Integer),
+        ),
+        # mypy expects keys in the dict to be type string, but the result of the df.to_dict() will have keys with type Hashable.
+        # This is known data, so not worrying about it.
+        orcid_stats,  # type: ignore
+    )
+
+
+def downgrade():
```
lwrubel, line 51: This downgrade does not leave any data there. But it doesn't seem like we need to preserve the limited version of the historic data, and can just reload.

Issue #562, opened 2025-09-15
Handle empty orcidid as null when loading authors CSV into author_orcids


lwrubel:
When loading `authors_active.csv` into the `author_orcids` table in the reports database, an empty orcidid needs to be NULL in the database. This supports the dashboard logic in Tableau that is looking for NULL or present ORCIDs. 

https://github.com/sul-dlss/rialto-airflow/blob/2b4bc9cc198feeee042490935d4ff9e5bf0d3511/rialto_airflow/publish/orcid.py#L37

PR #571, opened 2025-09-16
Handle empty orcidid as null when loading authors CSV into author_orcids

jmartin-sul:
closes https://github.com/sul-dlss/rialto-airflow/issues/562

### DIFF
--- a/rialto_airflow/publish/orcid.py
+++ b/rialto_airflow/publish/orcid.py
@@ -34,7 +34,7 @@ def export_author_orcids(data_dir):
                 row_values = {
                     "sunetid": row["sunetid"],
                     "full_name": row["full_name"],
-                    "orcidid": row["orcidid"],
+                    "orcidid": row["orcidid"] if row["orcidid"].strip() != "" else None,
                     "orcid_update_scope": row["orcid_update_scope"].lower() == "true",
                     "role": row["role"],
                     "primary_affiliation": row["primary_affiliation"],
--- a/test/publish/test_orcid.py
+++ b/test/publish/test_orcid.py
@@ -94,6 +94,27 @@ def authors_active_csv(tmp_path):
                 "true",
             ]
         )
+        writer.writerow(
+            [
+                "senaj",
+                "Enaj",
+                "Stanford",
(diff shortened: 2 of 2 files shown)


rialto_airflow/publish/orcid.py
```diff
@@ -34,7 +34,7 @@ def export_author_orcids(data_dir):
                 row_values = {
                     "sunetid": row["sunetid"],
                     "full_name": row["full_name"],
-                    "orcidid": row["orcidid"],
+                    "orcidid": row["orcidid"] if row["orcidid"].strip() != "" else None,
```
jmartin-sul, line 37: in practice, when i queried stage, i only saw empty string (`"""`, and got no results for whitespace (e.g. `orcidid = ''` got results, `orcidid LIKE ' %'` didn't).  but the whitespace stripping felt like a harmless bit of extra protection.

test/publish/test_orcid.py
```diff
@@ -109,8 +130,8 @@ def test_export_author_orcids(
     assert result is True
 
     with test_reports_session.begin() as session:
-        rows = session.query(AuthorOrcids).all()
-        assert len(rows) == 1
+        rows = session.query(AuthorOrcids).order_by(AuthorOrcids.sunetid).all()
```
jmartin-sul, line 133: added the sort since query results aren't guaranteed to be deterministically ordered otherwise, and flappy tests are no fun

test/publish/test_orcid.py
```diff
@@ -120,6 +141,8 @@ def test_export_author_orcids(
         assert rows[0].primary_school == "School of Engineering"
         assert rows[0].primary_department == "Computer Science"
         assert rows[0].primary_division == "Philanthropy Division"
+        assert rows[1].sunetid == "senaj"
+        assert rows[1].orcidid is None
```
jmartin-sul, line 145: happy to flesh this out and make the check of the second user more thorough.  i just concentrated on whether we were looking at the right record, and whether it had the expected value for the field whose behavior is modified by this PR.

lwrubel (approved):

PR #589, opened 2025-09-23
Move types generation into distill

### LINKED ISSUE #585: Understand why we have many missing publication types
In the `rialto_reports` database for a recent harvest there are 205,361 publications (25%) that lack a value in `publications.types`:

```SQL
select count(*) from publications
where types is null;

205361
```

Do these publications really not have a type? Do our rules for extracting the type need to be adjusted?

Is there a mapping problem ?  It is missing in the source data?

edsu:
Previously we were doing some complicated logic in SQL and Python to generate the list of publication types for our reporting. This commit moves it into the distill step instead, which results in a new column "types" on the "publication" table, which can be null or a list of string values.

In addition we are now extracting publication types for sulpub, crossref and pubmed ... so we shouldn't see null values anymore. There will be new values however. As discussed in standup (2025-09-24) we talked about follow up work to normalize these values during distill.

Closes #585


rialto_airflow/utils.py
```diff
@@ -57,27 +58,19 @@ def normalize_orcid(orcid):
...
-    if row.wos_type:
-        if isinstance(row.wos_type, list):
-            for wos_type in row.wos_type:
-                types.add(wos_type.lower())
-        else:
-            types.add(row.wos_type.lower())
-
-    return sorted(types)
-
-
```
edsu, line 76: This goes away since it is replace by our types distill rules.

jmartin-sul (approved):

Issue #593, opened 2025-09-25
alert loudly when the database migration container fails to exit cleanly (i.e. alert on non-zero exit code)


jmartin-sul:
i only thought to look at the prod logs because i saw this locally when spinning up my rialto-airflow docker stack: https://github.com/sul-dlss/rialto-airflow/issues/592

it'd be nice if we either saw a clear error message in the capistrano output (on the terminal of the developer doing the deployment) or if we got a honeybadger alert.

[discussed](https://stanfordlib.slack.com/archives/C0227EGGBFB/p1758830138582149?thread_ts=1758828507.209949&cid=C0227EGGBFB) with @edsu yesterday.  summary of our plan is that we will have other airflow containers depend on a successful exit by rialto-db-init.  if it fails, the webserver and worker containers should fail to start, `docker compose up` will fail, and then cap deploy will fail since it restarts the docker stack synchronously as one of its tasks.

PR #599, opened 2025-09-26
Depend on rialto-db-init contianer's successful exit where needed

jmartin-sul:
`airflow-webserver` and `airflow-worker` containers require our DB migrations to succeed, because they may write to our custom tables.

If `rialto-db-init` fails to exit cleanly, that indicates migrations failed.  Dependent containers failing to start in this situation has two advantages:
* `docker compose up` should fail, thus failing the capistrano deployment (since it restarts the docker stack), thus making it clear that there was an issue with deployment.
* Nagios healthcheck for the service (yet to be added) should fail, since the webserver will be down.

Since rialto-airflow is a data harvesting pipeline that runs periodically, the possible brief downtime implied by the above should be tolerable (and preferable to migrations failing silently on deploy).  The person deploying should notice that the deploy failed, and can at worst revert to a working commit while migration issues are debugged.

I confirmed that this PR will cause both laptop `docker compose up` and capistrano deployment to fail if the DB init container fails to complete sucessfully.  See https://github.com/sul-dlss/rialto-airflow/pull/598 for gory testing details.

I explicitly added the `rialto-db-init` dependency to the two containers in question in addition to the `airflow-init` dependency, even though `rialto-db-init` already depends on `airflow-init`.  This is because our team norm tends to be explicit specification of dependencies in package management configs, even if the dependency is implicitly/transitively available, for both clarity and stability (in the face of indirect dependency changes).  This seemed pretty analogous to that.

closes https://github.com/sul-dlss/rialto-airflow/issues/593

### DIFF
--- a/compose.prod.yaml
+++ b/compose.prod.yaml
@@ -158,6 +158,8 @@ services:
       <<: *airflow-common-depends-on
       airflow-init:
         condition: service_completed_successfully
+      rialto-db-init:
+        condition: service_completed_successfully
 
   airflow-scheduler:
     <<: *airflow-common
@@ -197,6 +199,8 @@ services:
     depends_on:
       <<: *airflow-common-depends-on
       airflow-init:
+        condition: service_completed_successfully
+      rialto-db-init:
         condition: service_completed_successfully
 
   airflow-triggerer:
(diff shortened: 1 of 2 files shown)


peetucket (approved):

edsu:
+1 the idea of retaining the `airflow-init` dependency and adding `rialto-db-init`. If `airflow-init` were ever removed or changed we would still want the airflow services to wait for `airflow-init` to finish.

Issue #609, opened 2025-10-08
Get publisher name from journal_issn


jacobthill:
OpenAlex has a sources API. Once we have the journal_issn data (dependent on https://github.com/sul-dlss/rialto-airflow/issues/586), we can use it to query this API to get a clean publisher name. We should also consider what other data we might want to get e.g. journal citation counts.

`url = f"https://api.openalex.org/sources/issn:{issn}`

lwrubel:
Oops, yes, you're right! Can we use `primary_location.source.host_organization_name`? The idea is to pull it when available from the data we already have rather than duplicate it with a call to the Source API.

lwrubel:
Based on querying the prod db publication table, when the DAG next runs, we should get a publisher from `openalex_json` for 465,674 records (51.6%). 

If/when we proceed with querying OpenAlex by ISSN for source data, it would be most productive to start with sulpub. Of the records with `sulpub_json` but no `openalex_json`, 193,795 have an issn populated. 

Of the other records with null `openalex_json` (without looking deeper for host_organization_name), the next highest is`wos_json` with 59,353 records and then `crossref` with 32,841, `pubmed` with `20,613`. `dim_json` only has 7,224.

PR #650, opened 2025-10-22
Get publisher from openalex_json

lwrubel:
Partly implements #609 to get the publisher name from the openalex_json when it is present as a `primary_location.source.host_organization_name`.

Tested locally.


edsu (approved):
This looks like a good thing to add as we discuss how to visualize the data.

Issue #632, opened 2025-10-14
Review harvest DAG logging


lwrubel:
Some harvest tasks are generating logs that are too large to access via the Airflow UI or download. This came up recently with the wos_harvest task. We've been successfully running most of these tasks so the amount of logging may no longer be necessary. Review the logging and pare down.

lwrubel:
Also implement a clean-up strategy (or spin off into another ticket if it ends up not fitting here).

jmartin-sul:
slack conversation about logs growing without bound on VM: https://stanfordlib.slack.com/archives/C09M7P91R/p1761067773423139

jmartin-sul:
> before closing this i'll:
> - [x] let #656 run in prod, to confirm that it reduces the amount of log output
> - [x] file an ops task ticket to deal with log retention

done?
- kam added a cron job (defined in puppet) for log cleanup.  
- i browsed the logs for the latest prod harvest DAG run (from yesterday).  logs loaded for all tasks, speedily for most, and seemed reasonably scrollable at a glance.  the `wos_harvest` logs still loaded slowly, and had this warning at the top: "Large log file. Some lines have been truncated. Download logs in order to see everything."  but this is still an improvement -- before this work, trying to load that task's logs would timeout and fail.  i was also able to download the entire (465 MB!) `wos_harvest` log from that same run.  this is also an improvement, since even trying to download the `wos_harvest` log failed for me before.

i'm going to close this for now, but if the WOS harvest log is still too unwieldy with this adjusted level of logging, we can reopen this ticket, or open a new follow on ticket specifically for `rialto_airflow.harvest.wos`.

PR #656, opened 2025-10-27
reduce log activity by adjusting log levels and consolidating some log statements

jmartin-sul:
part of #632 (maybe wait for a production run we're happy with to close?)

turn some info level statements to debug level, especially inside of loops, with the expectation that logging level in deployed envs will usually be set to info or warning.

turn a couple info level statements to warning level (esp around harvest limit setting used in test envs), so that they always show up in prod envs, where we woult typically not want limited harvests.

switching info statements to debug lets us keep them around in case they'd be useful again for troubleshooting, without adding noise in normal operation.

add a couple functions to test.utils for asserting expectations on log output.

### DIFF
--- a/rialto_airflow/distiller.py
+++ b/rialto_airflow/distiller.py
@@ -139,8 +139,17 @@ def _jsonpath_match(rule: JsonPathRule, data) -> Optional[str | int]:
         result = results[0].value
         if rule.only_positive_number:
             result = _ensure_positive_number(result)
+            if result is None:
+                logging.debug(
+                    f"could not cast '{results[0].value}' to a positive number; jpath={jpath}; data={data}"
+                )
         elif rule.is_valid_year:
             result = _ensure_valid_year(result)
+            if result is None:
+                logging.debug(
+                    f"could not cast '{results[0].value}' to a non-future year; jpath={jpath}; data={data}"
+                )
+
     else:
         result = None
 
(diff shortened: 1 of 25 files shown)


jmartin-sul:
there are only 3 uncovered lines, which i'll note inline as comments

rialto_airflow/distiller.py
```diff
@@ -152,10 +161,9 @@ def _ensure_positive_number(value: str | int) -> Optional[int]:
     try:
         result = int(value)
         if result < 0:
-            logging.warning(f"got a non-positive number {value}")
             result = None
     except (ValueError, TypeError):
-        logging.warning(f'got "{value}" instead of a positive number')
+        pass
```
peetucket: what does "pass" do?

rialto_airflow/funders/linker.py
```diff
@@ -45,8 +45,8 @@ def link_dim_publications(snapshot) -> int:
 
         for row in session.execute(stmt):
             count += 1
-            if count % 100 == 0:
-                logging.info(f"processed {count} publications from Dimensions")
+            if count % 50000 == 0:
+                logging.debug(f"processing publication number {count} from Dimensions")
```
jmartin-sul: as of 1 pm oct 28, this change from `info` to `debug` is uncovered, because the true case for the if is never exercised.

Issue #616, opened 2025-10-10
Populate publications_by_author with fields related to bibliographic metadata


lwrubel:
Publish the following fields to the publication_by_author table (add to `publish.publication.export_publications_by_author`) using the prioritization of data sources/fields described in [Shared Table Requirements ](https://docs.google.com/document/d/1tS0ZYnlqTj7TJ91OVh17srwXeJIGLs1srMJS5hQZRvs/edit?tab=t.0)

* abstract
* journal_issn
* journal_name
* pages
* issue
* title
* volume

lwrubel:
When this has been merged, deployed, and the publications_by_author table is populated, we should supply the data to Roop for feedback.

PR #660, opened 2025-10-30
Publish subset of bib metadata

lwrubel:
Partially completes #616. Populates the following fields in `publications_by_author`:

* journal_issn
* journal_name
* publisher
* title
* abstract
* pages

Run on stage as a test.

### DIFF
--- a/rialto_airflow/harvest/openalex.py
+++ b/rialto_airflow/harvest/openalex.py
@@ -1,9 +1,10 @@
+from functools import cache
 import json
 import logging
 import os
 from pathlib import Path
 
-from pyalex import Authors, Works, config
+from pyalex import Authors, Sources, Works, config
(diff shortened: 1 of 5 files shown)


rialto_airflow/harvest/openalex.py
```diff
@@ -195,4 +196,17 @@ def _clean_dois_for_query(dois: list[str]) -> list[str]:
 
 
 def _doi_log_message(doi: str):
-    logging.debug(f"dropping {doi} from openalex lookup")
+    logging.warning(f"dropping {doi} from openalex lookup")
+
+
+@cache
```
lwrubel: Queries OpenAlex sources for an ISSN when to get publisher data or a journal name when not already available in openalex_json.

PR #690, opened 2025-11-13
Ignore repeated sulpub records

### LINKED ISSUE #684: Prevent storing duplicate records from sulpub
As part of work on #630 we've noticed that our sulpub harvesting appears to be pulling in duplicate records (publications with the same sulpubid).

```sql
select sulpub_json -> 'sulpubid', count(*) from publication group by sulpub_json -> 'sulpubid' having count(*) > 1 order by count(*) desc;

 ?column? | count
----------+--------
 "166226" |      7
 "166177" |      7
 "166179" |      7
 "166202" |      7
 "171372" |      7
 "166184" |      7
 "166207" |      7
 "166173" |      7
 "166201" |      7
 "166186" |      7
 "166172" |      7
 "166178" |      7
 "166200" |      7
 "166183" |      7
 "166180" |      7
 "166191" |      7
 "166210" |      7
 "166190" |      7
 "166175" |      7
 "166192" |      7
 "166216" |      7
 "166217" |      7
 "166204" |      7
 "166209" |      7
 "166205" |      7
 "166219" |      7
 "166222" |      7
 "166225" |      7
...
```

It appears that there are 62,555 `sulpubid` values in the `publication.sulpub_json` that appear more than once.

```sql
select count(total) from (select sulpub_json -> 'sulpubid', count(*) as total from publication group by sulpub_json -> 'sulpubid' having count(*) > 1);

 count
-------
 62555
(1 row)
```

Our use of the sulpub API is pretty straightforward in that it just keeps a simple page count while fetching each page:

https://github.com/sul-dlss/rialto-airflow/blob/bc12fe2c4c50996c7d1aba8d9995c5905ef255c2/rialto_airflow/harvest/sul_pub.py#L68-L104

Rather than trying to understand and fix sulpub the easiest thing for us to do here would be to maintain a set of sulpubid values as we harvest, and to prevent storing the same publication more than once in the database.

Alternatively we could add a deduplication by `sulpubid` to our existing deduplication task?

edsu:
The sulpub API seems to (sometimes) return the same publication record more than once. This commit adds some logic to keep track of the sulpubid in each record, and filters out ones it has seen before.

Fixes #684


rialto_airflow/harvest/sul_pub.py
```diff
@@ -101,6 +102,14 @@ def publications(host, key, per_page=1000, limit=None):
                 more = False
                 break
 
+            # guard against the same record coming back twice
+            # see: https://github.com/sul-dlss/rialto-airflow/issues/684
+
+            sulpub_id = record["sulpubid"]
+            if sulpub_id in sulpub_ids:
+                continue
+            sulpub_ids.add(sulpub_id)
```
peetucket: is there any danger that this runs into memory problems with a very large set of ids?

edsu: I used the `sulpub.jsonl` from the most recent crawl to extract all the `sulpubid`s. This created a set of 455,685 unique values (out of the 726,236 total with duplicates). The memory footprint for the set looks like it is 36 MB, so I think we'll be OK?

The downside to doing it this way is that if we were to restart the sulpub harvesting task it would lose the memory of having already stored some of the publications. I'm going to separately explore how we could add sulpub to the deduplication task. Doing it there would mean the harvest task could be restarted any number of times.

PR #694, opened 2025-11-15
Distill refactor

edsu:
Previous to this PR the rules for extracting metadata resided in different places: `rialto_airflow.harvest.distill` and `rialto_airflow.publish.publication`. All the rules and their dependent functions were also in single modules, which got a bit unwieldy for finding and editing.

This commit consolidates the rules into the `rialto_airflow.distiller` module. The related tests were also moved to `test/distill`. In many cases the tests were simplified to just test the specific distill function, without requiring setting up and tearing down a database. One test was left in `test/harvest/distill.py` to ensure that the `rialto_airflow.harvest.distill` still works.


lwrubel (approved):
Assuming the results of running this on stage are OK, I am fine with merging this and running in prod.

PR #708, opened 2025-11-25
Adding natural key deduplication

### LINKED ISSUE #685: Implement natural key based deduplication for publications
After work on #630 we decided to implement a natural key based approach to folding together what appear to be duplicate publications that lack a DOI from the `publication` table. To do this we need to:

- rename existing `remove_duplicates` Airflow task to `remove_platform_duplicates`
- add a new Airflow `remove_publication_duplicates` task that follows `distill` and precedes `link_funders`
- the new task needs to:
  - find publications that lack a DOI, and have a title, pub_year, publisher, journal_name and first_author_name
  - generate key for said publications (using technique from the [notebook](https://github.com/sul-dlss-labs/rialto-notebooks/blob/main/publication-duplication.ipynb)
  - group together publications with the same key
  - for each group select a publication to fold into (the one with the most platform json metadata)
  - add any missing metadata from other publications to the selected publication
  - update the publication-authors linkages similar to https://github.com/sul-dlss/rialto-airflow/blob/c8309924d34590f8ff93c9c18c74161865b91438/rialto_airflow/harvest/deduplicate.py#L119-L139 
  - delete the orphaned publications
- the process should be generous with logging so we can see what it is doing and get some counts at the end.

edsu:
This commit adds a step to the deduplication Airflow task to fold together publications that have similar title, publication year and first author.

Closes #685

### DIFF
--- a/rialto_airflow/distiller/__init__.py
+++ b/rialto_airflow/distiller/__init__.py
@@ -15,6 +15,7 @@ from .volume import volume
 from .citation_count import citation_count
 from .author_names import author_list_names, first_author_name, last_author_name
 from .author_orcids import author_list_orcids, first_author_orcid, last_author_orcid
+from .natural_key import natural_key
 
 
 __all__ = [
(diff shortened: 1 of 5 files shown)


edsu:
I think this is very stale at this point.

PR #717, opened 2025-12-03
Change name of role column in author table

lwrubel:
Previously, we were taking the `authors.csv` "role" column and writing to a "primary_role" column in the harvest database `Author` table. Since rialto-orgs and the authors.csv file also have a "primary_role" column (which we are not currently using) this was confusing. The purpose of this change is to make it clearer which data we are using from rialto-orgs. 

Tested by running a harvest locally.


peetucket (approved):
Remind me how the current database column name is going to get renamed?  We don't need a migration for this?

lwrubel:
The harvest database does not have migrations. We have a schema and a new database is created with each harvest.

PR #732, opened 2025-12-12
cli utility for rialto

edsu:
This is the start of a cli tool for rialto database. It will generate a CSV of publications for a given sunet. We could expand it to do other things though?

```
uv run rialto publications makeller
```

or to list the authors:

```
uv run rialto authors
```

There was no pressing need for the `authors` subcommand, but if there was only one subcommand typer just made it the default command, and I wanted it to be `publications`.

### DIFF
--- a/rialto_airflow/cli.py
+++ b/rialto_airflow/cli.py
@@ -1,0 +1,107 @@
+import os
+import csv
+import sys
+from pathlib import Path
+
+import dotenv
+import typer
+from typing_extensions import Annotated
+from sqlalchemy import select
+
+from rialto_airflow.snapshot import Snapshot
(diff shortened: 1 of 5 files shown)


rialto_airflow/cli.py
```diff
@@ -0,0 +1,103 @@
...
+                }
+            )
+
+
+@app.command()
+def authors(data_dir="data", db_name: Annotated[str, typer.Option()] = ""):
+    """
+    List the SUNET IDs for authors in the database.
+    """
+    if db_name is None:
```
lwrubel: I think this needs to be 
```suggestion
    if db_name == "":
```

lwrubel (approved):

PR #743, opened 2025-12-17
cli can use remote db

edsu:
This commit adds:

- a `--db-uri` option to connect to a remote database
- a `snapshots` command to list snapshot databases available
- output platform identifiers for tracking down data if necessary


edsu:
This is too stale now, and some of the code is no longer relevant.

PR #755, opened 2026-01-09
Log fill-in JSON-L files separately

edsu:
To use the JSON-L harvest logs in our study of ORCID reliability  it is helpful to be able to distinguish between what was collected by ORCID, and what was collected as part of the "fill in" process by DOI.

This commit separates out the harvest and fill-in JSON-L files. It also creates an "envelope" for data that is harvested by ORCID that includes the ORCID that was used. So harvested data that looked like this:

```json
{
  "id": 123
}
```

will now look like:

```json
{
  "orcid": "0000-0002-1825-0097", 
  "data": {
    "id": 123
  }
}
```

Separating out the fill in JSON-L means that we can remove some JSON-L mocks that were testing that the JSON-L file was being appended to.

I ran this in my development environment with a limit of 1000 and it seemed to work fine, but led me to noticing #757

### DIFF
--- a/rialto_airflow/harvest/pubmed.py
+++ b/rialto_airflow/harvest/pubmed.py
@@ -3,6 +3,7 @@ import logging
 import os
 import re
 from pathlib import Path
+from typing import Any
 
 import requests
 import xmltodict
@@ -18,7 +19,7 @@ from rialto_airflow.schema.harvest import (
     pub_author_association,
 )
 from rialto_airflow.snapshot import Snapshot
-from rialto_airflow.utils import normalize_doi
+from rialto_airflow.utils import normalize_doi, add_orcid
 
 Params = Dict[str, Union[int, str]]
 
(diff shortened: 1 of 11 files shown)


rialto_airflow/harvest/pubmed.py
```diff
@@ -173,7 +176,7 @@ def pmids_from_dois(dois: list[str]) -> list[str]:
     return _pubmed_search_api(batch_query)
 
 
-def publications_from_pmids(pmids: list[str]) -> list[str]:
+def publications_from_pmids(pmids: list[str]) -> list[dict[Any, Any]]:
```
edsu: mypy noticed this type annotation was wrong when I was writing it to the jsonl file with `add_orcid()` that had types...

test/harvest/test_dimensions.py
```diff
@@ -250,27 +249,6 @@ def test_log_message(snapshot, mock_authors, mock_many_dimensions, caplog):
...
-            "doi": "10.1515/9781503624151",
-            "title": "Another example title",
-            "publication_year": 1892,
-        },
-    ]
-    with open(path, "w") as f:
-        for record in records:
-            f.write(f"{json.dumps(record)}\n")
-
-
```
edsu: The testing got a bit easier since we no longer need to mock a harvested JSON-L file when testing the fill-in process. Fill in will write to its own JSON-L file. You will see this repeated below.

Issue #765, opened 2026-01-20
fill_in_wos not writing data


edsu:
It looks like the `fill_in_wos` task is running, and talking to the Web of Science API, but not writing anything to the `wos.jsonl` or the publications database?

The Airflow task log look like a lot of messages like this:

```
[...]

PR #771, opened 2026-01-20
Speed up WoS fill-in

edsu:
This commit updates the Wos fill-in process to wait less time between HTTP 429 requests. We were noticing that our fill_in_wos Airflow task was taking a long time (multiple days) and never completed. The sleep time between requests was too big, when requesting metadata one DOI at a time. The time was lowered to be no longer than 1 minute on the 10th retry.

Also this change ensures that the DOI is extracted correctly from the WoS metadata when determining what publication to update in the database. Apparently the fill-in process for WoS has never been updating the database because we weren't getting the DOI out of the metadata correctly. This only became apparent once we started logging the fill-in JSON-L metadata separate from the harvest.

Refs #765

### DIFF
--- a/test/harvest/test_wos.py
+++ b/test/harvest/test_wos.py
@@ -95,10 +95,18 @@ def mock_wos_doi(monkeypatch):
 
     def f(*args, **kwargs):
         yield {
-            "doi": "10.1515/9781503624199",
             "title": "An example title",
             "type": "article",
             "publication_year": 1891,
+            "dynamic_data": {
+                "cluster_related": {
+                    "identifiers": {
+                        "identifier": [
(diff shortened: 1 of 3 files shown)


test/harvest/test_wos.py
```diff
@@ -95,10 +95,18 @@ def mock_wos_doi(monkeypatch):
...
+            "dynamic_data": {
+                "cluster_related": {
+                    "identifiers": {
+                        "identifier": [
+                            {"type": "doi", "value": "10.1515/9781503624199"}
+                        ]
+                    }
+                }
+            },
         }
```
edsu: Adjust the mock to reflect where the DOI is *actually* found in WoS metadata.

PR #785, opened 2026-02-06
Harvest Dimensions concepts

edsu:
Concepts were useful to have for the h3 team, and shouldn't flake out the Dimensions API hopefully? We had pared back what we were requesting a few weeks ago because the API became very unreliable when we requested every field that was possible. It seems like Dimensions concepts would be useful to have for RIALTO as well.


lwrubel (approved):
🤞

PR #799, opened 2026-03-02
Small refactor in WoS tests

peetucket:
~~The WoS API appears to have changed.  It takes a database parameter, and we have been using a value of "WOK", which is supposed to be "all databases").  However, the API now appears to be returning fewer results with "WOK" that it does with a parameter of "WOS" (which is their primary database, the Web of Science Expanded API).  This is unexpected, since WOK is supposed to be ALL databases, including the WOS database.  I will contact them about this, since it will also affect sul-pub.  I verified it has nothing to do with our RIALTO API key (I verified same issues with sul-pub key, WOK now has fewer results than WOS databases).~~

The above is now all resolved.  But a small refactor to the test to make it clear is left here for consideration.  No change in functionality.

### DIFF
--- a/rialto_airflow/harvest/wos.py
+++ b/rialto_airflow/harvest/wos.py
@@ -219,7 +219,7 @@ def _wos_api(
     count = 100
 
     params: Params = {
-        "databaseId": "WOK",
+        "databaseId": "WOS",
         "usrQuery": query,
         "count": count,
         "firstRecord": 1,
--- a/rialto_airflow/harvest/wos.py
+++ b/rialto_airflow/harvest/wos.py
@@ -219,7 +219,7 @@ def _wos_api(
     count = 100
 
     params: Params = {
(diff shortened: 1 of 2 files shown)


test/distiller/test_publisher.py
```diff
@@ -94,7 +94,7 @@ def test_pubmed_publisher_journal(pubmed_json):
     assert journal_issn(pub) == "1873-2054"
     # will do live lookup in OpenAlex Sources API
     assert journal_name(pub) == "Health & Place"
-    assert publisher(pub) == "Elsevier"
+    assert publisher(pub) == "Elsevier BV"
```
jmartin-sul: this line will have a minor conflict to resolve with https://github.com/sul-dlss/rialto-airflow/pull/798, which reproduces the same change to get the dep update build passing for the week

PR #807, opened 2026-04-01
prefer dimensions to openalex for open access info

### LINKED ISSUE #796: Prefer Dimensions to Open Alex for Open Access category
We are currently getting Preprints from OpenAlex that are designated `Closed` for their Open Access category. Some background on this can be found in this ticket: https://github.com/sul-dlss/rialto-tableau/issues/82

A spreadsheet of these 952 publications can be found in https://docs.google.com/spreadsheets/d/1k4iVRdpqGjvXWfqp8XTKYcnKQBF5wOVlAKnnC9jlmRc/edit?usp=sharing

The proposal is to adjust our [Open Access rules](https://github.com/sul-dlss/rialto-airflow/blob/main/rialto_airflow/distiller/open_access.py) so that we prefer Dimensions to OpenAlex where possible for Open Access category since Dimensions had these categorized as `Green` instead of `Closed` for 568 of the 952.

One open question is whether we should ensure that we never have Open Access `Closed` for `Preprints`. Examining the spreadsheet should hopefully help resolve that.

peetucket:
Fixes #796

Work related to ensuring we don't have Preprints that are Closed was split off to #808

### DIFF
--- a/test/distiller/test_open_access.py
+++ b/test/distiller/test_open_access.py
@@ -4,7 +4,7 @@ from rialto_airflow.harvest.distill import open_access
 
 def test_dimensions(dim_json):
     """
-    open_access should come from dimensions if unavailable in openalex
+    open_access should come from dimensions
     """
     pub = Publication(doi="10.1515/9781503624153", dim_json=dim_json)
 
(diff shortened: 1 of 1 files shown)


test/distiller/test_open_access.py
```diff
@@ -2,22 +2,9 @@
...
-    """
-    pub = Publication(
-        doi="10.1515/9781503624153",
-        openalex_json=openalex_json,
-        dim_json=dim_json,
-    )
-
-    assert open_access(pub) == "gold"
-
-
```
peetucket: this test is basically the same as one below (the "fallback to openalex")

test/distiller/test_open_access.py
```diff
@@ -26,7 +13,7 @@ def test_dimensions(dim_json):
 
 def test_open_access_null(dim_json):
     """
-    dimensions is still used when there is an empty in openalex open_access
+    open_access should come from dimensions when openalex is available but has no open_access info
```
edsu: This comment is misleading. We always prefer Dimensions to Open Alex, whether it is there or not. Maybe the test should ensure that a None value in the dimensions metadata causes it to fall back to Open Alex?

PR #810, opened 2026-04-03
Airflow 3

### LINKED ISSUE #482: Upgrade to Airflow v3
We are currently Airflow 2.10, installed as part of the Docker base image. 

Upgrading docs: https://airflow.apache.org/docs/apache-airflow/stable/installation/upgrading_to_airflow3.html

edsu:
Note: This code was used to successfully run a harvest in stage with a limit of 50,000.

These are changes that were required to upgrade to Airflow 3. This involved changes to our Docker configuration to enable new services. It also included an upgrade of sqlalchemy to version 2, which required a few changes, and also allowed us to re-enable type checking in several places.

Another additional change is that Airflow 3 removed the ability to pickle objects in XCOM. pickling was deemed to be a security risk. We were relying on pickling to pass the Snapshot object around, which contains a pathlib.Path object, which is not serializable. So a custom XCom backend was created which allowed Snapshot objects to be JSON serializable.

Note: the docker configuration still needs to be tested in stage, but works in development.

Fixes #482

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -47,7 +47,7 @@ jobs:
           AIRFLOW_VAR_RIALTO_POSTGRES: "postgresql+psycopg2://airflow:airflow@localhost"
 
       - name: Run type checking
-        run: uv run mypy .
+        run: uv run ty check .
 
       - name: Run tests
         run: uv run pytest --cov-branch --cov-report=xml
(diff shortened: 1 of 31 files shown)


lwrubel (approved):
Approved pending publish_to_reports DAG completing successfully on stage.

PR #831, opened 2026-04-10
Add fill-in task for altmetric

### LINKED ISSUE #375: Altmetric
The Altmetric API lets you collect information about the impact of research by DOI. We could add a workflow task to collect it and store it in a new column on the publication table.

From @tomcramer

edsu:
*Ready for review, but marked as hold until it has been tested in stage*

---

**What was changed?**

This adds a fill-in task for fetching information from Altmetric for every publication that has a DOI.

It uses the "Fetch" API endpoint: https://docs.altmetric.com/details-page-api/endpoints/fetch/

The data can include social media (twitter, bluesky, linkedin, reddit, etc) as well as more traditional web news sources.

Before merging an altmetric key will need to be added to Vault, and Puppet updated to put it in the environment at `AIRFLOW_VAR_ALTMETRIC_KEY`.

Closes #375

- Need to do this if/when merged


edsu:
Running this in stage for a few hours we eventually got stopped because we only retry HTTP 429 errors 10 times. I adjusted the logging to include the text of the final response and the headers to indicate what the problem is.

> ERROR    root:altmetric.py:80 Altmetric rate limit (429): body={"error":"Tweet ID rate limit exceeded, please see X-TweetIdRateLimit headers and try again later."} limit_headers={'X-HourlyRateLimit-Limit': '3600', 'X-HourlyRateLimit-Remaining': '3600', 'X-DailyRateLimit-Limit': '86400', 'X-DailyRateLimit-Remaining': '86400', 'X-TweetIDRateLimit-Limit': '1500000', 'X-TweetIDRateLimit-Remaining': '0'}

It looks like there is a limit on how many tweet IDs you can get in a day. There is also a limit of 86,400 requests in general per day. This means that it would take us 7.5 days to do the full harvest of data (646,095 DOIs), afterwhich the incrementals would work quickly. More about their rate limiting can be read [here](https://docs.altmetric.com/details-page-api/getting-started/#rate-limiting) (although it doesn't document the actual limits)

I'm leaning towards not introducing this kind of delay into our full harvest even if we aren't doing it that often. The fill-in process (which this Altmetric data collection is a part of) current runs for 24 hours for a full harvest.

I will reach out to our Dimensions/Altmetric contact and see if there's a way of upping our daily limit. But I suspect they will probably push us in the direction of using their [Google BigQuery](https://docs.altmetric.com/bigquery/), which might actually be [doable](https://airflow.apache.org/docs/apache-airflow-providers-google/stable/operators/cloud/bigquery.html) in our Airflow workflow...

PR #837, opened 2026-04-16
Dedupe by dimensions id

### LINKED ISSUE #504: Deduplicate by Dimensions id where DOI is null
Background: There are currently only 694 rows in the publications table which have an `dim_json` value but a null `doi`. 56 of those ids have duplicate publications (143 publication rows total). This is much fewer than the other platforms. 

Add a step to deduplicate on the "id" key in the `dim_json` following the approach in https://github.com/sul-dlss/rialto-airflow/pull/495.

Recommend working on this after #503 and following any refactoring or abstracting that happens with work with that issue.

lwrubel:
**What was changed?**
Closed #504 to dedupe by Dimensions ID when the DOI is null. I considered making a separate non-unique column for dim_json but wasn't sure if that was needed at this point for a within-platform dedupe. Open to adding that, though. 

Added note about deduplication.


edsu (approved):

edsu (approved):
Nice, yeah we might as well keep the logic the same for now, instead of adding a dimensions_id column.

PR #849, opened 2026-04-17
Add type annotations to doi functions

### LINKED ISSUE #698: Type annotations for normalize_doi()
I think it would be helpful to add type annotations to the `normalize_doi()` function (and its internal functions). The internal functions should also be named with a `_` prefix to indicate they aren't for use outside of the module?

lwrubel:
**What was changed?**
Closes #698 to add type annotations.


lwrubel:
I'm not sure why Codecov is flagging a path as being uncovered. It's covered in two tests including the one I added to ensure this is covered.

peetucket (approved):

PR #855, opened 2026-04-20
Set up incremental harvest DAG as copy of harvest

### LINKED ISSUE #840: Create harvest2 DAG
To facilitate work on incremental harvesting without disturbing our ability to run the current harvesting workflow we are going to create a `harvest2` DAG that is identical to the existing [`harvest` DAG](https://github.com/sul-dlss/rialto-airflow/blob/main/rialto_airflow/dags/harvest.py). We should not schedule it to run automatically.

Since we will be modifying the harvesting code a lot we should also copy the harvesting functions and tests as well.

- [ ] `cp rialto_airflow/dags/harvest.py rialto_airflow/dags/harvest2.py`
- [ ] `cp -r rialto_airflow/harvest rialto_airflow/harvest2`
- [ ] `cp -r tests/harvest tests/harvest2`
- [ ] update imports in harvest2 DAG and copied tests to import from `rialto_airflow/dags/harvest2`.

We may need to silence some type warnings about similarly named functions / classes, but maybe not.

lwrubel:
**What was changed?**
Closes #840 to create a DAG for incremental harvest, set up side-by-side with the existing harvest DAG, and simply as a copy of the code to start. 

Note that the issue specified calling the DAG `harvest2`, but I called it `harvest_incremental` here. Open to changing it back to `harvest2` if that's preferred.

### DIFF
--- a/rialto_airflow/dags/harvest_incremental.py
+++ b/rialto_airflow/dags/harvest_incremental.py
@@ -1,0 +1,210 @@
+import datetime
+import logging
+from pathlib import Path
+import shutil
+
+from airflow.sdk import dag, task, task_group
+from airflow.models import Variable
+
+from rialto_airflow import funders
+from rialto_airflow.harvest_incremental import (
+    authors,
+    crossref,
+    dimensions,
+    openalex,
+    sul_pub,
+    wos,
+    pubmed,
+    distill,
+    deduplicate,
+)
+from rialto_airflow.database import create_database, create_schema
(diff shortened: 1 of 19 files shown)


peetucket (approved):

edsu:
Nice! I prefer `harvest_incremental` since we can probably keep that after harvest goes away,  without the bigger of renaming back to harvest.

edsu (approved):
This looks good! I propose we ignore the test coverage warning.

PR #858, opened 2026-04-22
Set up database for incremental harvest and tests

### LINKED ISSUE #841: Database migration for `rialto`
Currently we are creating a new database every time the harvest runs. As part of incremental harvesting work we will want to reuse the same database every week. We need to add a migration for the current schema we are using for harvesting. Unless it causes problems we could still keep `rialto_reports` database distinct and call this new database simply `rialto`. At some point we may want to fold them together to simplify things?

We should update the `rialto_airflow.database.get_session()` function so that when a database name is not passed in it returns a session for the `rialto` database. Or if it's easier we could create a new function that does that. If the latter proves easier that will require some changes to how the code in `rialto_airflow.harvest2` get their session.

The `rialto_airflow.dag.harvest2` setup task will need to be adjusted to not create the database, and `rialto_airflow.harvest2` code will need to be adjusted to not pass in a database name to `get_session()` or the new function (if there is one).

lwrubel:
**What was changed?**
Fixes #841. This sets up a separate schema for the incremental harvesting database, which I think will make it easier to add columns or make changes along the way. It does however require setting up a separate test database (`rialto_incremental_test`) from the existing `rialto_test` database and some separate parallel test fixtures.


rialto_airflow/schema/rialto.py
```diff
@@ -0,0 +1,129 @@
...
+    declarative_base,
+    mapped_column,
+    relationship,
+)
+from sqlalchemy.types import DateTime
+
+from rialto_airflow.database import utcnow
+
+# permanent database for incrementally harvested data
+RIALTO_DB_NAME: str = "rialto"
```
lwrubel, line 16: Rather than change `get_session()` to pass in none, this sets up a constant with the database name, to be used in other harvesting. It helps for getting the right database name to the separate test database for the harvesting as well.

alembic_rialto/versions/53aa35d55767_initialize_rialto_db.py
```diff
@@ -0,0 +1,355 @@
...
+    op.drop_table("author_orcids")
+    op.drop_table("publications_by_department")
+    op.drop_table("publications")
+    # ### end Alembic commands ###
+
+
+def downgrade() -> None:
+    """Downgrade schema."""
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.create_table(
```
edsu: I thought that the downgrade would just drop all the tables? Is there a reason why it is creating all the reports tables here?

rialto_airflow/schema/rialto.py
edsu: I like this simplification!

edsu (changes requested):
I left one question about the downgrade Alembic step, which seemed to contain some things I didn't expect to see.

I noticed for the first time that the alembic commands in the README.md are slightly off:

`uv run dotenv alembic -n rialto upgrade head`

needs to be:

`uv run dotenv run alembic -n rialto upgrade head`

I only noticed this because bringing up the containers with `docker compose up` which in turns runs the `rialto-db-init` container, didn't seem to apply the migrations for the `rialto` database. I do seem to have a `rialto` database, but it has no tables in it.

Maybe this was user error?

alembic_rialto/versions/53aa35d55767_initialize_rialto_db.py
lwrubel: I've cleaned up the migrations and tested by: dropping the rialto database, rebuilding the docker containers when bringing them up, and am then seeing the correct tables. 

```
airflow=# \c rialto
psql (16.6 (Homebrew), server 16.13 (Debian 16.13-1.pgdg13+1))
You are now connected to database "rialto" as user "airflow".
rialto=# \dt
                 List of relations
 Schema |          Name          | Type  |  Owner  
--------+------------------------+-------+---------
 public | alembic_version        | table | airflow
 public | author                 | table | airflow
 public | funder                 | table | airflow
 public | pub_author_association | table | airflow
 public | pub_funder_association | table | airflow
 public | publication            | table | airflow
(6 rows)
```

edsu (approved):
Yay, awesome to see this working.

PR #861, opened 2026-04-22
use mais orcid prod API for consistency in development

### LINKED ISSUE #421: Fix inconsistency with MaIS API configs
I just noticed that we have MaIS API URLs hardcoded to prod in the docker compose.yml file (so that it matches the compose.prod.yml file and allows us to run the publish_orcid DAG in stage against the actual MaIS data)....but that the MaIS tests expect to hit UAT MaIS data.    This means that if try to run the publish_orcid DAG in localhost, you need to have the prod MaIS API keys in your .env file, but if you want to run the tests, you need the UAT MaIS API keys in your .env file (as described in the README).  This is kind of annoying, either we need to use have stage hit the MaIS UAT API, or we need to adjust our tests to use prod.  I'll write a ticket so we can decide what to do at some point.

See https://github.com/sul-dlss/rialto-airflow/blob/main/README.md?plain=1#L171-L180
and https://github.com/sul-dlss/rialto-airflow/blob/main/compose.yaml#L89-L90 and https://github.com/sul-dlss/rialto-airflow/blob/main/test/test_mais.py#L86-L91

peetucket:
**What was changed?**

Fixes #421  Allows us to test against the Mais ORCID Prod API, instead of UAT.  This allows us to run a local harvest against prod data without having to reconfigure our local .env just to run the tests.

In order to complete this work, we need to: 

- [x] update env values for Mais API for the client_id and client_secret in github secrets to use prod values (CI will fail until we do this)
- [ ] remind devs to update their local .env file to use prod credentials (else local MaIS tests will fail)
- [x] remove vault values from `puppet/application/rialto-airflow/test`

### DIFF
--- a/rialto_airflow/mais.py
+++ b/rialto_airflow/mais.py
@@ -121,7 +121,11 @@ def fetch_orcid_user(access_token: str, base_url: str, user_id: str) -> ORCIDRec
 
 
 def current_orcid_users(
-    mais_client_id: str, mais_client_secret: str, mais_token_url: str, base_url: str
+    mais_client_id: str,
+    mais_client_secret: str,
+    mais_token_url: str,
+    base_url: str,
+    limit: Optional[int] = None,
 ) -> list[ORCIDRecord]:
     """Retrieves the current ORCID records from the MAIS ORCID API."""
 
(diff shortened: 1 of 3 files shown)


rialto_airflow/mais.py
```diff
@@ -121,15 +121,19 @@ def fetch_orcid_user(access_token: str, base_url: str, user_id: str) -> ORCIDRec
 
 
 def current_orcid_users(
-    mais_client_id: str, mais_client_secret: str, mais_token_url: str, base_url: str
+    mais_client_id: str,
+    mais_client_secret: str,
+    mais_token_url: str,
+    base_url: str,
+    limit: Optional[int] = None,
```
peetucket, line 128: allowing an optional limit on this method makes testing easier

lwrubel (approved):

PR #862, opened 2026-04-23
preprints have no apc cost

### LINKED ISSUE #797: APC for Preprints should be 0
In looking at [an anomaly in our Open Access visualizations](https://github.com/sul-dlss/rialto-tableau/issues/82) that have publications of type `Preprint` with Open Access status:

```
select open_access, count(*) from publication where 'Preprint' = any(types) group by open_access; open_access | count
-------------+-------
 bronze      |     4
 closed      |  1960
 diamond     |    43
 gold        | 26682
 green       | 33348
 hybrid      |     7
             |    21
```

When we don't find an APC value in our metadata we are [assigning one](https://github.com/sul-dlss/rialto-airflow/blob/7cbe1dc0c36b06449f60afa3dabd413f5d372790/rialto_airflow/distiller/apc.py#L27-L38) based on Open Access status:

diamond: $0
gold: 2450
hybrid: 3600

This means that we are accumulating $65,370,900 in APC value for Gold Preprints and arguably there should be no APC charge?

One way to solve for this would be to adjust our [APC rules](https://github.com/sul-dlss/rialto-airflow/blob/main/rialto_airflow/distiller/apc.py) so that we don't return an APC value for Preprints.

peetucket:
**What was changed?**

Fixes #797 

Note: gemini helped me find how to check if a publication is a preprint and in writing the spec

### DIFF
--- a/rialto_airflow/distiller/apc.py
+++ b/rialto_airflow/distiller/apc.py
@@ -7,7 +7,7 @@ def apc(pub, context):
     Get the APC cost from one place in the openalex data, an external dataset, or another place in
     OpenAlex data.
     """
-    if "Preprint" in (context.get("types") or []):
+    if context.get("types") == ["Preprint"]:
         return 0
 
     # https://dataverse.harvard.edu/dataset.xhtml?persistentId=doi:10.7910/DVN/CR1MMV
--- a/rialto_airflow/distiller/apc.py
+++ b/rialto_airflow/distiller/apc.py
@@ -7,6 +7,9 @@ def apc(pub, context):
     Get the APC cost from one place in the openalex data, an external dataset, or another place in
     OpenAlex data.
     """
+    if "Preprint" in (context.get("types") or []):
+        return 0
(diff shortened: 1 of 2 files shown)


edsu (approved):
short n' sweet!

Just saying this out loud. If a Publication has multiple types like `['Preprint', 'Article']` we will assign $0 APC. Is this what we want? Or should its only type be `Preprint`?

The good news is that it looks it is pretty rare for Preprints to be of any other type:

```
select types, count(*) from publication where 'Preprint' = any(types) group by types order by count(*) desc;

       types        | count
--------------------+-------
 {Preprint}         | 63963
 {Article,Preprint} |     4
```

peetucket:
Good point, maybe it should be exclusively pre-print (even though it is rare).

peetucket:
Tweaked PR to only set to 0 if preprint is the only type (and adjusted test)

edsu (approved):

PR #864, opened 2026-04-27
Create Harvest class for harvest_incremental DAG

### LINKED ISSUE #842: Create and use Harvest class
We need a `Harvest` ORM model for recording when a harvest starts, and when it finishes. The harvest2 DAG should create a Harvest object, and pass it to each of the harvest tasks. The Harvest class should have a `get_previous()` method that looks in the database for that most recent completed Harvest. 

We are no longer going to be saving the JSON-lines harvest metadata. So we do not need the equivalent of `Snapshot.path`.

- [ ] create database model
- [ ] remove creation of Snapshot from `riallto_airflow.dags.harvest2.setup` 
- [ ] add creation of a Harvest to `riallto_airflow.dags.harvest2.setup` 
- [ ] pass Harvest to each of the harvest DAG tasks in `riallto_airflow.dags.harvest2`
- [ ] update `rialto_airflow.harvest2` functions to accept `Harvest` instead of `Snapshot`
- [ ] remove all JSON-lines writing from `rialto_airflow.harvest2` module.

lwrubel:
**What was changed?**
Fixes #842 to add a Harvest class, remove jsonl file-writing and Snapshot usage from the harvest_incremental DAG. 

I ran this locally to confirm the table was added when the rialto-db-init container was built.


rialto_airflow/dags/harvest.py
```diff
@@ -181,7 +181,7 @@ def link_funders(snapshot):
         """
         Link all the publications to funders.
         """
-        funders.link_publications(snapshot)
+        funders.link_publications(snapshot.database_name)
```
lwrubel, line 184: Since the `funders` module must temporarily support both the `harvest` and `harvest_incremental` DAGs, it needs to not require a Snapshot to exist.

rialto_airflow/harvest_incremental/harvests.py
```diff
@@ -0,0 +1,46 @@
+import datetime
+
+from sqlalchemy import select
+
+from rialto_airflow.database import get_session
+from rialto_airflow.schema.rialto import Harvest, RIALTO_DB_NAME
+
+
+def create_harvest() -> Harvest:
```
lwrubel: These are helper methods for the `Harvest` class that has been added to the `rialto` database.

rialto_airflow/harvest_incremental/harvests.py
```diff
@@ -0,0 +1,46 @@
...
+
+def get_harvest_by_id(harvest_id: int) -> Harvest | None:
+    with get_session(RIALTO_DB_NAME).begin() as session:
+        harvest = session.get(Harvest, harvest_id)
+        if harvest is not None:
+            session.expunge(harvest)
+        return harvest
+
+
+def get_previous() -> Harvest | None:
```
lwrubel: Not used yet, but was requested in the ticket, I think for later incremental harvest logic?

rialto_airflow/harvest_incremental/harvests.py
```diff
@@ -0,0 +1,46 @@
+import datetime
```
lwrubel: Calling this module `harvests` rather than `harvest` to avoid confusion and naming conflicts with the existing `harvest` DAG and `harvest` schema. We could consider renaming this later, once that code is removed, if it seems worthwhile.

rialto_airflow/dags/harvest_incremental.py
```diff
@@ -56,150 +55,148 @@ def harvest_incremental():
...
+        pubmed_harvest(harvest_id)
 
     @task()
-    def fill_in_openalex(snapshot):
+    def fill_in_openalex(harvest_id):
         """
         Fill in OpenAlex data for DOIs from other publication sources.
         """
-        openalex.fill_in(snapshot)
+        openalex.fill_in()
```
edsu, line 117: So future PRs we will modify these harvest functions and pass in the `harvest_id`?

rialto_airflow/dags/harvest_incremental.py
```diff
@@ -56,150 +55,148 @@ def harvest_incremental():
...
         """
-        Set up the snapshot directory and database.
+        Create a Harvest record to track this run.
         """
-        snapshot = Snapshot.create(data_dir)
-        shutil.copyfile(Path(rialto_authors_file(data_dir)), snapshot.authors_csv)
-
-        return snapshot
+        harvest = harvests.create_harvest()
+        return harvest.id
```
edsu, line 60: I like the idea of passing around the `harvest_id` rather than making the Harvest class JSON serializable for XCom.

rialto_airflow/harvest_incremental/deduplicate.py
```diff
@@ -8,27 +8,19 @@
...
-        + dimensions_dupes
-        + sulpub_dupes
-        + wos_id_dupes
-        + pubmed_id_dupes
-    )
+    openalex_dupes = remove_openalex_duplicates()
+    dimensions_dupes = remove_dimensions_duplicates()
+    sulpub_dupes = remove_sulpub_duplicates()
+    wos_id_dupes = remove_wos_id_duplicates()
+    pubmed_id_dupes = remove_pubmed_id_duplicates()
```
edsu, line 23: Same question here, we will add back `harvest_id` to these?

rialto_airflow/harvest_incremental/harvests.py
lwrubel: The session logic (using `flush()` and `expunge()`) was suggested by Copilot, but seems to make sense to avoid problems with how Harvest information is passed around.

rialto_airflow/harvest_incremental/harvests.py
edsu: Does it make sense to put these functions in the Harvest ORM class itself as class methods?

So you could use them like: 

```python
h = Harvest.get_latest(session)
```

edsu (approved):
Wow, this was a big set of changes! I left a question about whether it might make sense to add helper methods to the Harvest class rather than a new harvests module. But this looks like a good way to move forward!

lwrubel:
I had originally put the helper methods for `Harvest` in the class definition in `schema.rialto` but I wasn't sure if that was a helpful pattern or not, so moved it out to its own module. But it sounds like it would make more sense to do that based on your reading, so I'll refactor that.

rialto_airflow/harvest_incremental/deduplicate.py
lwrubel: I think so, once we have something to do with it.

rialto_airflow/dags/harvest_incremental.py
lwrubel: Phew!

lwrubel:
@edsu I refactored to move the helper methods and added some test coverage for that module. Github seems to be cooperating again. Would you take a look and confirm all is OK before merging?

test/harvest_incremental/test_harvest.py
```diff
@@ -0,0 +1,74 @@
...
+@pytest.fixture
+def mock_rialto_db_name(monkeypatch):
+    monkeypatch.setattr(rialto, "RIALTO_DB_NAME", "rialto_incremental_test")
+
+
+def test_create_persists_harvest(test_incremental_session, mock_rialto_db_name):
+    harvest = Harvest.create()
+
+    assert harvest.id is not None
+    assert inspect(harvest).detached is True
```
lwrubel, line 19: Checking that the session is not active and that `expunge()` worked as expected.

edsu (approved):

PR #868, opened 2026-04-28
Add _harvested columns to publication table

### LINKED ISSUE #859: Add provenance columns to publication table
We would like to know where publications came from originally (i.e. at harvest time, not fill-in time).

Add one column per source that holds a timestamp.  The appropriate timestamp will be added/updated when the publication row is created by a harvester (not by a filein).

e.g. wos_harvested, pubmed_harvested, openalex_harvested, dimensions_harvested

lwrubel:
**What was changed?**
Fixes #859 to track harvest source. Columns will be NULL until populated in the harvest tasks.

Tested locally to confirm they start out as NULL.

N/A


alembic_rialto/versions/ad881a79ea4d_add_provenance_columns.py
```diff
@@ -0,0 +1,81 @@
...
+        ),
+    )
+    op.add_column(
+        "publication",
+        sa.Column(
+            "pubmed_harvested",
+            sa.DateTime(),
+            server_default=sa.text("TIMEZONE('utc', CURRENT_TIMESTAMP)"),
+            nullable=True,
+        ),
```
edsu, line 63: I guess the reason to not to add `crossref_harvested` is because we aren't harvesting it by ORCID, but filling it in after the fact?

edsu (approved):
Looks good!

alembic_rialto/versions/ad881a79ea4d_add_provenance_columns.py
lwrubel: Right, since we're tracking provenance from harvest by ORCID.

PR #869, opened 2026-04-28
Harvest dimensions incrementally

### LINKED ISSUE #844: Dimensions incremental harvest
Update the `rialto_airflow.harvest2.dimensions.harvest()` function to use the `Harvest.previous()` method to determine when the last successful harvest *started*. Use that time to fetch records that have been updated since. If no previous Harvest is found then fetch all publications.

lwrubel:
**What was changed?**
Closes #844. Implements incremental harvesting from Dimensions. Populates the `dim_harvested` field with the current time for provenance. 

Supports using a harvest limit for publications and authors, and harvests all pubs for new/updated authors. 

The publications and authors don't change. We'll want to update the changelog with something about incremental harvesting when this is released.

### DIFF
--- a/rialto_airflow/harvest_incremental/dimensions.py
+++ b/rialto_airflow/harvest_incremental/dimensions.py
@@ -36,8 +36,7 @@ def harvest(limit: None | int = None) -> None:
     with get_session(RIALTO_DB_NAME).begin() as select_session:
         previous_harvest = Harvest.get_previous()
         previous_harvest_date = None
-        if previous_harvest is not None and previous_harvest.created_at is not None:
-            # Dimensions expects the date in YYYY-MM-DD format.
+        if previous_harvest is not None:
             previous_harvest_date = previous_harvest.created_at.strftime("%Y-%m-%d")
 
         # get all authors that have an ORCID
--- a/rialto_airflow/harvest_incremental/dimensions.py
+++ b/rialto_airflow/harvest_incremental/dimensions.py
@@ -33,6 +33,12 @@ def harvest(limit: None | int = None) -> None:
     stop = False
 
     with get_session(RIALTO_DB_NAME).begin() as select_session:
+        previous_harvest = Harvest.get_previous()
+        previous_harvest_date = None
+        if previous_harvest is not None and previous_harvest.created_at is not None:
+            # Dimensions expects the date in YYYY-MM-DD format.
+            previous_harvest_date = previous_harvest.created_at.strftime("%Y-%m-%d")
+
         # get all authors that have an ORCID
         for author in (
             select_session.query(Author).where(Author.orcid.is_not(None)).all()
@@ -41,7 +47,9 @@ def harvest(limit: None | int = None) -> None:
                 logging.warning(f"Reached limit of {limit} publications stopping")
                 break
 
-            for dimensions_pub_json in publications_from_orcid(author.orcid):
+            for dimensions_pub_json in publications_from_orcid(
+                author.orcid, harvest_date=previous_harvest_date
+            ):
                 count += 1
                 if limit is not None and count > limit:
                     stop = True
@@ -103,23 +111,14 @@ def publications_from_dois(dois: list, batch_size=200):
             yield normalize_publication(pub)
 
 
-def publications_from_orcid(orcid: str, batch_size=200):
+def publications_from_orcid(orcid: str, batch_size=200, harvest_date=None):
     """
     Get the publications metadata for a given ORCID, for records created (inserted) since the last harvest.
     """
-    previous_harvest = Harvest.get_previous()
-    # Dimensions expects the date in YYYY-MM-DD format.
-    previous_harvest_date = None
-    if previous_harvest is not None and previous_harvest.created_at is not None:
-        previous_harvest_date = previous_harvest.created_at.strftime("%Y-%m-%d")
     orcid = normalize_orcid(orcid)
     logging.debug(f"looking up publications for orcid {orcid}")
     fields = " + ".join(publication_fields())
-    date_limit = (
-        f' and date_inserted >= "{previous_harvest_date}"'
-        if previous_harvest_date
-        else ""
-    )
+    date_limit = f' and date_inserted >= "{harvest_date}"' if harvest_date else ""
     q = f"""
         search publications where researchers.orcid_id = "{orcid}"{date_limit}
         return publications[{fields}]
--- a/rialto_airflow/harvest_incremental/dimensions.py
+++ b/rialto_airflow/harvest_incremental/dimensions.py
@@ -29,7 +29,8 @@ def harvest(limit: None | int = None) -> None:
     """
     Walk through all the Author ORCIDs and generate publications for them.
     """
-    count = 0
+    pub_count = 0
+    author_count = 0
     stop = False
 
     with get_session(RIALTO_DB_NAME).begin() as select_session:
@@ -43,15 +44,33 @@ def harvest(limit: None | int = None) -> None:
         for author in (
             select_session.query(Author).where(Author.orcid.is_not(None)).all()
         ):
+            author_count += 1
+            if limit is not None and author_count > limit:
+                stop = True
+
             if stop is True:
-                logging.warning(f"Reached limit of {limit} publications stopping")
+                logging.warning(
+                    f"Reached limit of {limit} publications or authors, stopping"
+                )
                 break
+
+            # if the author was created or updated after the last harvest (e.g. adding an ORCID),
+            # we want to get all their publications, not just ones since the last harvest timestamp.
+            if previous_harvest is not None:
+                # TODO: remove the created_at check once updated_at gets populated upon create
+                if author.created_at >= previous_harvest.created_at:
+                    previous_harvest_date = None
+                elif (
+                    author.updated_at is not None
+                    and author.updated_at >= previous_harvest.created_at
+                ):
+                    previous_harvest_date = None
 
             for dimensions_pub_json in publications_from_orcid(
                 author.orcid, harvest_date=previous_harvest_date
             ):
-                count += 1
-                if limit is not None and count > limit:
+                pub_count += 1
+                if limit is not None and pub_count > limit:
                     stop = True
                     break
 
--- a/rialto_airflow/harvest_incremental/dimensions.py
+++ b/rialto_airflow/harvest_incremental/dimensions.py
@@ -1,3 +1,4 @@
+import datetime
 import logging
 import os
 import time
@@ -10,12 +11,13 @@ from sqlalchemy import select, update
 from sqlalchemy.dialects.postgresql import insert
 
 from rialto_airflow.database import get_session
-from rialto_airflow.schema.harvest import (
+from rialto_airflow.schema.rialto import (
+    Harvest,
     Author,
     Publication,
     pub_author_association,
+    RIALTO_DB_NAME,
 )
-from rialto_airflow.schema.rialto import RIALTO_DB_NAME
 from rialto_airflow.utils import (
     normalize_doi,
     normalize_orcid,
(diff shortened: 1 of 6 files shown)


rialto_airflow/schema/rialto.py
```diff
@@ -66,15 +66,17 @@ class Publication(RialtoSchemaBase):
...
     wos_id: Mapped[Optional[str]] = mapped_column(String)
     pubmed_id: Mapped[Optional[str]] = mapped_column(String)
-    openalex_harvested: Mapped[Optional[DateTime]] = mapped_column(DateTime)
-    dim_harvested: Mapped[Optional[DateTime]] = mapped_column(DateTime)
-    sulpub_harvested: Mapped[Optional[DateTime]] = mapped_column(DateTime)
-    wos_harvested: Mapped[Optional[DateTime]] = mapped_column(DateTime)
-    pubmed_harvested: Mapped[Optional[DateTime]] = mapped_column(DateTime)
-    created_at: Mapped[Optional[DateTime]] = mapped_column(
+    openalex_harvested: Mapped[Optional[datetime.datetime]] = mapped_column(DateTime)
+    dim_harvested: Mapped[Optional[datetime.datetime]] = mapped_column(DateTime)
```
lwrubel, line 70: Correcting the type annotations to a Python datetime object, not a SQL Alchemy DateTime. This came up when type checking data to be added to the field.

test/test_database.py
```diff
@@ -93,12 +93,7 @@ def test_drop_database(
             teardown_database(snapshot.database_name)
 
 
-def test_create_schema(
-    snapshot,
-    mock_rialto_postgres,
-    monkeypatch,
-    teardown_database,
-):
+def test_create_schema(snapshot, mock_rialto_postgres, monkeypatch, teardown_database):
```
lwrubel, line 96: Just a formatting change, not sure what changed it.

rialto_airflow/harvest_incremental/dimensions.py
```diff
@@ -99,16 +105,25 @@ def publications_from_dois(dois: list, batch_size=200):
...
     """
+    previous_harvest = Harvest.get_previous()
+    # Dimensions expects the date in YYYY-MM-DD format.
+    previous_harvest_date = None
+    if previous_harvest is not None and previous_harvest.created_at is not None:
+        previous_harvest_date = previous_harvest.created_at.strftime("%Y-%m-%d")
     orcid = normalize_orcid(orcid)
     logging.debug(f"looking up publications for orcid {orcid}")
     fields = " + ".join(publication_fields())
+    date_limit = (
```
lwrubel: `date_inserted` is the query param and needs to be YYYY-MM-DD format. Only add a date range if there is a previously completed harvest.

rialto_airflow/harvest_incremental/dimensions.py
```diff
@@ -27,21 +29,48 @@ def harvest(limit: None | int = None) -> None:
...
+        previous_harvest_date = None
+        if previous_harvest is not None and previous_harvest.created_at is not None:
+            # Dimensions expects the date in YYYY-MM-DD format.
+            previous_harvest_date = previous_harvest.created_at.strftime("%Y-%m-%d")
+
         # get all authors that have an ORCID
         for author in (
             select_session.query(Author).where(Author.orcid.is_not(None)).all()
         ):
+            author_count += 1
```
lwrubel, line 46: When running an incremental harvest in a development setup with a harvest limit, it is useful to also limit the number of authors looked up, since adding the date limit in harvesting could make it so that all of the authors (or a large number) would have to be queried.

rialto_airflow/harvest_incremental/dimensions.py
```diff
@@ -27,21 +29,48 @@ def harvest(limit: None | int = None) -> None:
...
                 break
 
-            for dimensions_pub_json in publications_from_orcid(author.orcid):
-                count += 1
-                if limit is not None and count > limit:
+            # if the author was created or updated after the last harvest (e.g. adding an ORCID),
+            # we want to get all their publications, not just ones since the last harvest timestamp.
+            if previous_harvest is not None:
+                # TODO: remove the created_at check once updated_at gets populated upon create
+                if author.created_at >= previous_harvest.created_at:
```
lwrubel, line 60: Start checking for new/updated authors since it's important we don't put a date limit on querying for their publications the first time they're in our system.

rialto_airflow/harvest_incremental/dimensions.py
```diff
@@ -27,21 +29,48 @@ def harvest(limit: None | int = None) -> None:
...
     """
-    count = 0
+    pub_count = 0
+    author_count = 0
     stop = False
 
     with get_session(RIALTO_DB_NAME).begin() as select_session:
+        previous_harvest = Harvest.get_previous()
+        previous_harvest_date = None
+        if previous_harvest is not None and previous_harvest.created_at is not None:
```
edsu: If previous_harvest is not None we can't rely on previous_harvest.created_at always being set. I guess this was `ty` not being able to reason about the sqlalchemy Harvest model?

rialto_airflow/schema/rialto.py
edsu: Weird, I wonder why we didn't hit these warnings before. I didn't notice them when testing updated_at, created_at in my load_authors work.

test/conftest.py
```diff
@@ -71,6 +73,12 @@ def test_session(test_engine):
         close_all_sessions()
 
 
+@pytest.fixture
+def mock_rialto_db_name(monkeypatch):
+    monkeypatch.setattr(rialto, "RIALTO_DB_NAME", "rialto_incremental_test")
+    monkeypatch.setattr(dimensions, "RIALTO_DB_NAME", "rialto_incremental_test")
+
```
edsu, line 80: I'm curious why this change wasn't needed before.

test/harvest_incremental/test_dimensions.py
```diff
@@ -35,115 +70,186 @@ def test_publication_fields():
...
 
+    previous_harvest = Harvest.get_previous()
+    assert previous_harvest is not None
+    assert previous_harvest.created_at is not None
 
-@pytest.fixture
-def mock_rialto_db_name(monkeypatch):
-    monkeypatch.setattr(dimensions, "RIALTO_DB_NAME", "rialto_incremental_test")
+    previous_harvest_date = previous_harvest.created_at.strftime("%Y-%m-%d")
+    assert previous_harvest_date == "2026-04-27", "date formatted as YYYY-MM-DD"
```
edsu: I'm a bit confused what is being tested here. It seems to be testing that sqlalchemy can persist dates correctly?

test/harvest_incremental/test_dimensions.py
```diff
@@ -35,115 +70,186 @@ def test_publication_fields():
...
+    monkeypatch,
+):
+    with test_incremental_session.begin() as session:
+        session.add(
+            Harvest(
+                created_at=datetime.datetime(2026, 4, 27, 16, 38, 10),
+                finished_at=datetime.datetime(2026, 4, 28, 0, 0, 0),
+            )
+        )
+        # Make sure Authors are similar to those loaded in productio, with created_at dates.
```
edsu: small typo `productio`

test/harvest_incremental/test_dimensions.py
```diff
@@ -35,115 +70,186 @@ def test_publication_fields():
...
-        patched_dsl.query_iterative
-    )  # a ref to the real query_iterative function
+    harvest_dates = []
 
-    # for the first call to query_iterative that does an orcid query on publications,
-    # raise a request exception.  for the rest, just call the real query function.
-    req_count = 0
+    def _capture_harvest_date(orcid, harvest_date=None):
+        harvest_dates.append(harvest_date)
+        return iter(())
```
edsu: `yield` instead of `return iter(())` seems a bit more readable to me.

test/harvest_incremental/test_dimensions.py
```diff
@@ -35,115 +70,186 @@ def test_publication_fields():
...
+        # set of publications from the API.
+        session.add(
+            Harvest(
+                created_at=datetime.datetime(2010, 4, 27, 16, 38, 10),
+                finished_at=datetime.datetime(2010, 4, 28, 0, 0, 0),
+            )
+        )
+    previous_harvest = Harvest.get_previous()
+    if previous_harvest is not None and previous_harvest.created_at is not None:
+        previous_harvest_date = previous_harvest.created_at.strftime("%Y-%m-%d")
```
edsu: I think we are testing get_previous() elsewhere so this can be:

```python
assert previous_harvest is not None
assert previous_harvest.created_at is not None
```

or not :-)

edsu (approved):
I left a few suggestions and questions, but no blockers!

There are some creative approaches to testing that the harvest dates are getting used when querying Dimensions.

rialto_airflow/harvest_incremental/dimensions.py
lwrubel: Yes, exactly. I didn't start with this logic, but since there was theoretically a chance it could be None, then `strftime()` wouldn't work so `ty` was unhappy.

rialto_airflow/schema/rialto.py
lwrubel: The warnings were based on trying to do `strftime()` on a previous harvest created_at date. But it seems if I add an ignore rule line around that logic, I can simplify the `None` checking and also change these back to `DateTime` across the board. It seems safe enough to do since there will always be a  `created_at` date.

rialto_airflow/harvest_incremental/dimensions.py
lwrubel: Going to ignore this rule in this one spot since it seems safe to do so and also allows us to not change the type annotations in the the rialto schema.

rialto_airflow/schema/rialto.py
edsu: I'm ok with the changes, as long as that's what people are generally doing with SQLAlchemy.

lwrubel: Based on the docs (https://docs.sqlalchemy.org/en/20/orm/declarative_tables.html#mapped-column-derives-the-datatype-and-nullability-from-the-mapped-annotation) it looks like a Python datatype gets mapped to a SQLAlchemy type. Which makes me lean toward keeping the changes in the schema, to support type checking  of any operations we're doing on dates.

lwrubel: And if we then set `created_at: Mapped[datetime.datetime]` rather than `Mapped[Optional[datetime.datetime]]`, which I think is accurate, then I don't have to override the rule and can use simpler logic in working with a previous harvest date (not all those `None` checks). Will push this up so you can review.

rialto_airflow/harvest_incremental/dimensions.py
lwrubel: See comment above about adjusting the type annotations in the rialto schema and then not having to check for None.

test/conftest.py
lwrubel: This monkeypatch is in the other test files for incremental harvesting, so I thought it would be better to start adding this to `conftest.py` instead of in the individual test files. It's really `monkeypatch.setattr(rialto, "RIALTO_DB_NAME", "rialto_incremental_test")` that benefits from not being replicated everywhere. Maybe it makes more sense to monkeypatch `(dimensions, "RIALTO_DB_NAME", "rialto_incremental_test")` only in `test_dimensions.py` since I think that's the only place that will need it?

test/harvest_incremental/test_dimensions.py
lwrubel: That makes sense. Updated and added some comments.

test/harvest_incremental/test_dimensions.py
lwrubel: I made another pass through the tests to try to streamline them a little bit. Dropped this one in favor of the other harvest date checking.

rialto_airflow/schema/rialto.py
edsu: Nice!

PR #870, opened 2026-04-29
Incrementally load authors

### LINKED ISSUE #843: load_authors DAG task should "upsert" authors
The `rialto_airflow.harvest2.authors.load_authors` function should *update* information in the `author` table based on the information in the CSV. There is a primary key constraint on `sunetid` which can use a `on_conflict_do_update` clause similar to what we do when harvesting publications that share a DOI, e.g.

https://github.com/sul-dlss/rialto-airflow/blob/b2a422b5ac4271911a0a39ada49a31dfa3b9e564/rialto_airflow/harvest/pubmed.py#L102-L105

We should add a test to ensure that the update is working as expected.

edsu:
**What was changed?**

To support incremental updates we want to be able to reload `authors.csv` into an already populated author table.

The author table has three unique constraints: 

1. sunet: If we get a unique constraint violation for sunet, we updated the existing row with the potentially new information.
2. cap_profile_id: If we get a unique constraint violation for the cap_profile_id, we look for an inactive author, and if found, update it with the new author's information.
3. orcid: If we get a unique constraint violation for the orcid we log it and move on.

Care must also be taken to ensure that we only update Author.updated_at when data actually changes.

The task output log message on completion indicating how many authors were created, updated or skipped.

In addition to the unit tests, I ran the load_authors task in my development environment multiple times to ensure that it behaved correctly.

Aside: the task runs in 2 minutes instead of 15 now that we aren't starting a new session for each row.

Closes #843

**Notable data changes?**  Update the [Data Changelog](https://github.com/sul-dlss/rialto-web/wiki/Data-Changelog) ✅

### DIFF
--- a/rialto_airflow/harvest_incremental/authors.py
+++ b/rialto_airflow/harvest_incremental/authors.py
@@ -1,11 +1,13 @@
 import csv
 import logging
 
+from psycopg2 import Error as Psycopg2Error
+from sqlalchemy import update
 from sqlalchemy.exc import IntegrityError
-from sqlalchemy.orm import sessionmaker
+from sqlalchemy.orm import Session
 
-from rialto_airflow.database import get_engine
-from rialto_airflow.schema.rialto import Author, RIALTO_DB_NAME
+from rialto_airflow.database import get_session
+from rialto_airflow.schema.rialto import RIALTO_DB_NAME, Author
 from rialto_airflow.utils import rialto_authors_file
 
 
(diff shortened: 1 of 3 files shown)


rialto_airflow/harvest_incremental/authors.py
edsu, line 1: It might be easier to read the new file whole than to read the diff because so much has changed in `authors.py`.

rialto_airflow/harvest_incremental/authors.py
```diff
@@ -13,49 +17,135 @@ def load_authors_table(data_dir) -> None:
...
+    # is_distinct_from is a NULL-safe inequality check (unlike !=)
+    #
+    # the changed clause allows our update to only change rows that have
+    # actually changed, since we only want updated_at to get updated if
+    # something has changed in the row
+    column_changed = [
+        getattr(Author, col).is_distinct_from(getattr(stmt.excluded, col))
+        for col in row_dict
+    ]
+    changed = or_(*column_changed)
```
edsu: hopefully the comment adequately describes what is going on here? without this changed clause we would always set updated_at to the current time, which would make it useless for harvesting to use as a indicator of what needs a full harvest

rialto_airflow/schema/rialto.py
```diff
@@ -74,7 +74,9 @@ class Publication(RialtoSchemaBase):
     created_at: Mapped[Optional[DateTime]] = mapped_column(
         DateTime, server_default=utcnow()
     )
-    updated_at: Mapped[Optional[DateTime]] = mapped_column(DateTime, onupdate=utcnow())
+    updated_at: Mapped[Optional[DateTime]] = mapped_column(
+        DateTime, default=utcnow(), onupdate=utcnow()
+    )
```
edsu, line 79: The `default` parameter is needed to set `updated_at` when inserting the new row. I don't think a db migration is needed for this kind of change.

test/harvest_incremental/test_authors.py
```diff
@@ -34,51 +57,30 @@ def test_author_fixture(test_incremental_session, author):
...
+                "academic_council": "false",
+                "primary_affiliation": "Engineering",
+                "primary_school": "School of Engineering",
+                "primary_department": "Computer Science",
+                "primary_division": "Philanthropy Division",
+                "all_schools": "Independent Labs, Institutes, and Centers (Dean of Research)|School of Humanities and Sciences",
+                "all_departments": "Computer Science|Horticulture",
+                "all_divisions": "Philanthropy Division|Other Division",
+                "active": "true",
+            }
```
edsu, line 82: I switched the tests over to using a dict as a row, since it simplified some things, and made the column ordering less important to get right.

rialto_airflow/harvest_incremental/authors.py
```diff
@@ -1,61 +1,164 @@
...
+    # this prevents changing updated_at when there are no changes
+    if not any(getattr(author, col) != row_dict[col] for col in row_dict):
+        return "noop"
+
+    # update the author using the new data, and commit
+    for col, val in row_dict.items():
+        setattr(author, col, val)
+    session.add(author)
+    session.commit()
+    return "updated"
```
edsu, line 110: It simplified the logic to look for the user first, and have conditional logic based on the result, instead of trying to get fancy with an atomic Postgres upsert operation. Since we will only be running one `load_authors` process at a time I thought readability was more important than guarding against race conditions from someone else updating the same author at the same time.

rialto_airflow/schema/rialto.py
lwrubel: Thanks for figuring this out as part of the authors work.

rialto_airflow/harvest_incremental/authors.py
lwrubel: It makes this pretty readable and understandable too, thanks!

test/harvest_incremental/test_authors.py
```diff
@@ -108,7 +112,55 @@ def test_load_authors_table(
...
+
+def test_update_by_sunet(
+    test_incremental_session, tmp_path, caplog, authors_csv, mock_rialto_db_name
+):
+    """
+    Make sure that user updates work correctly. Reloading the same data should
+    not change the author.updated_at value. Any changes to user's data should
+    get loaded and update the author.updated_at timestamp.
+    """
+    # load initial user
```
lwrubel: Suggest using the term "author" instead of "user" throughout in comments to make it more consistent with the model / code and understandable.

rialto_airflow/harvest_incremental/authors.py
```diff
@@ -1,61 +1,164 @@
...
+        logging.warning(
+            f"Unable to insert active author when there is already an active author with cap_profile_id {cap_profile_id}"
+        )
+        return False
+
+    old_sunet = inactive_author.sunet
+
+    session.execute(
+        update(Author).where(Author.id == inactive_author.id).values(row_dict)
+    )
```
lwrubel, line 154: So by updating the existing author with that cap_profile_id, we retain the linkages with their publications (if any), nice.

lwrubel (approved):
Approved with a minor suggestion.

rialto_airflow/harvest_incremental/authors.py
edsu: Yes, I was deleting the user, and then reattaching the publications, but it seemed easier to just to update in place.

test/harvest_incremental/test_authors.py
edsu: Yes, thanks for noticing this. I kept confusing myself, and it helps to be consistent.

PR #874, opened 2026-05-01
Harvest updates from sul_pub

### LINKED ISSUE #848: sulpub incremental harvest
Update the `rialto_airflow.harvest2.sulpub.harvest()` function to use the `Harvest.previous()` method to determine when the last successful harvest *started*. Use that time to fetch records that have been updated since. If no previous Harvest is found then fetch all publications.

edsu:
**What was changed?**

This change will use the most recent harvest date as way to fetch only recently updated publications.

It includes a migration to ensure that deleting a Publication will cascade to the `pub_funder_association` table. I had a situation where the `remove_duplicates` step was trying to remove a publication and wasn't able to:

```
IntegrityError: (psycopg2.errors.ForeignKeyViolation) update or delete on table "publication" violates foreign key constraint "pub_funder_association_publication_id_fkey" on table "pub_funder_association"
DETAIL:  Key (id)=(20756) is still referenced from table "pub_funder_association".
```

I think this also means we should adjust the `remove_duplicates` task to preserve links between publications and funders, like it already does for authors. See #875 

Closes #848

n/a

### DIFF
--- a/alembic_rialto/versions/e1d3bfbd9ffd_cascade_delete_pub_funder_association.py
+++ b/alembic_rialto/versions/e1d3bfbd9ffd_cascade_delete_pub_funder_association.py
@@ -1,0 +1,51 @@
+"""cascade delete pub_funder_association
+
+Revision ID: e1d3bfbd9ffd
+Revises: ad881a79ea4d
+Create Date: 2026-05-01
+
+"""
+
+from typing import Sequence, Union
+
+from alembic import op
+
+
+# revision identifiers, used by Alembic.
+revision: str = "e1d3bfbd9ffd"
+down_revision: Union[str, Sequence[str], None] = "ad881a79ea4d"
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
+
+
+def upgrade() -> None:
+    """Upgrade schema."""
+    op.drop_constraint(
+        "pub_funder_association_publication_id_fkey",
+        "pub_funder_association",
+        type_="foreignkey",
+    )
+    op.create_foreign_key(
+        "pub_funder_association_publication_id_fkey",
+        "pub_funder_association",
+        "publication",
+        ["publication_id"],
+        ["id"],
+        ondelete="CASCADE",
+    )
+
+
+def downgrade() -> None:
+    """Downgrade schema."""
+    op.drop_constraint(
+        "pub_funder_association_publication_id_fkey",
+        "pub_funder_association",
+        type_="foreignkey",
+    )
+    op.create_foreign_key(
+        "pub_funder_association_publication_id_fkey",
+        "pub_funder_association",
+        "publication",
+        ["publication_id"],
+        ["id"],
+    )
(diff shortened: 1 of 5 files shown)


alembic_rialto/versions/e1d3bfbd9ffd_cascade_delete_pub_funder_association.py
edsu, line 1: This change is unrelated to the sulpub changes, but I ran into it while doing them. In testing locally I ended up with a Publication linked to a Funder which got marked as a duplicate during `remove_duplicates` task, which the database wouldn't allow because it would break the foreign key constraint. Setting Publication / Funder links to cascade like Publication / Author allows this to work. Note: the delete just cacscades to the join table, no Funders are removed.

alembic_rialto/versions/e1d3bfbd9ffd_cascade_delete_pub_funder_association.py
```diff
@@ -0,0 +1,51 @@
...
+from alembic import op
+
+
+# revision identifiers, used by Alembic.
+revision: str = "e1d3bfbd9ffd"
+down_revision: Union[str, Sequence[str], None] = "ad881a79ea4d"
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
+
+
```
edsu, line 20: Claude said this was the most reliable way to modify the `ondelete` behavior.

rialto_airflow/schema/rialto.py
edsu, line 1: This is the change to support the migration above.

test/harvest_incremental/test_sul_pub.py
```diff
@@ -1,28 +1,13 @@
...
-
-
-dotenv.load_dotenv()
-
-sul_pub_host = os.environ.get("AIRFLOW_VAR_SUL_PUB_HOST")
-sul_pub_key = os.environ.get("AIRFLOW_VAR_SUL_PUB_KEY")
-
-no_auth = not (sul_pub_host and sul_pub_key)
-
-
```
edsu, line 25: The tests don't actually use a mocked out sul_pub server since it's not available from CI. So this machinery was removed.

test/conftest.py
```diff
@@ -77,6 +77,7 @@ def test_session(test_engine):
 def mock_rialto_db_name(monkeypatch):
     monkeypatch.setattr(rialto, "RIALTO_DB_NAME", "rialto_incremental_test")
     monkeypatch.setattr(dimensions, "RIALTO_DB_NAME", "rialto_incremental_test")
+    monkeypatch.setattr(sul_pub, "RIALTO_DB_NAME", "rialto_incremental_test")
```
edsu, line 80: Hopefully when we are only using `harvest_incremental` these won't be needed...

test/conftest.py
```diff
@@ -231,13 +232,12 @@ def test_incremental_engine(monkeypatch):
...
 @pytest.fixture
-def test_incremental_session(test_incremental_engine, monkeypatch):
+def test_incremental_session(test_incremental_engine):
     """
     Returns a sqlalchemy session for the test database.
     """
-    # monkeypatch.setattr(publication, "RIALTO_DB_NAME", "rialto_incremental_test")
     try:
-        yield sessionmaker(engine_setup("rialto_incremental_test", echo=True))
+        yield sessionmaker(test_incremental_engine)
```
edsu, line 240: I aligned this with how `test_session()` works.

lwrubel (approved):

PR #876, opened 2026-05-03
Fix bug with Dimensions author harvest logic

lwrubel:
**What was changed?**
Fixes an issue with incremental harvesting where the previous harvest date for the first author was then used for the rest of the authors.


edsu (approved):
Nice catch - it's a bit hard to follow, but the comment helps.

lwrubel:
Is there a way I could make it less hard to follow? I'm not sure which part is the issue.

PR #879, opened 2026-05-04
Remove mock_rialto_db_name override

edsu:
**What was changed?**

I ran into an issue with the harvest_incremental sulpub test not using
the right database (rialto_test instead rialto_incrmental_test). The
problem turned out to be that the test module was overriding the
mock_rialto_db_name fixture.

I took the opportunity to import and separate out the different
sqlalchemy schemas in the same way, to make it a bit easier to digest.

There was also some unnecessary dotenv loading going on that can be
removed since that is happening in conftest.

n/a

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -43,23 +43,6 @@ addopts = "-v --cov --cov-report=html --cov-report=term --log-level INFO --log-f
 [tool.coverage.run]
 omit = ["test/*"]
 
-[tool.mypy]
-check_untyped_defs = true # Type-checks the interior of functions without type annotations.
-
-[[tool.mypy.overrides]]
-module = [
-    "dimcli",
-    "honeybadger",
-    "pyalex",
-    "requests_oauthlib",
-    "sqlalchemy_utils",
-    "jsonpath_ng",
-    "dominate",
-    "dominate.tags",
-    "typing_extensions"
-]
-ignore_missing_imports = true
-
 [dependency-groups]
 dev = [
     "apache-airflow==3.1.8",
--- a/rialto_airflow/harvest_incremental/__init__.py
+++ b/rialto_airflow/harvest_incremental/__init__.py
@@ -1,0 +1,23 @@
+from rialto_airflow.harvest_incremental import (
+    authors,
+    crossref,
+    deduplicate,
+    dimensions,
+    distill,
+    openalex,
+    pubmed,
+    sul_pub,
+    wos,
+)
+
+__all__ = [
+    "authors",
+    "crossref",
+    "deduplicate",
+    "dimensions",
+    "distill",
+    "openalex",
+    "pubmed",
+    "sul_pub",
+    "wos",
+]
--- a/test/conftest.py
+++ b/test/conftest.py
@@ -5,26 +5,13 @@ from sqlalchemy.orm import sessionmaker
 from sqlalchemy.orm.session import close_all_sessions
 from sqlalchemy_utils import create_database, database_exists, drop_database
 
+from rialto_airflow import harvest_incremental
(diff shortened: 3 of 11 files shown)


rialto_airflow/harvest_incremental/__init__.py
edsu, line 1: This was needed in order to avoid some type warnings in test/conftest.py

test/harvest_incremental/test_authors.py
```diff
@@ -44,11 +43,6 @@ def author(test_incremental_session):
         )
 
 
-@pytest.fixture
-def mock_rialto_db_name(monkeypatch):
-    monkeypatch.setattr(authors_mod, "RIALTO_DB_NAME", "rialto_incremental_test")
-
-
```
edsu, line 51: This fixture override interfered with the one we want to use in test/conftest.py There are a bunch more in other test modules below.

test/harvest_incremental/test_crossref.py
```diff
@@ -1,23 +1,13 @@
...
 from rialto_airflow.harvest_incremental import crossref
 from test.utils import num_log_record_matches
 
 
-@pytest.fixture
-def mock_rialto_db_name(monkeypatch):
-    monkeypatch.setattr(crossref, "RIALTO_DB_NAME", "rialto_incremental_test")
-
-
-dotenv.load_dotenv()
```
edsu, line 18: dotenv load happens once in test/conftest.py

test/conftest.py
```diff
@@ -5,26 +5,13 @@
...
-from rialto_airflow.schema.rialto import (
-    pub_author_association as rialto_pub_author_association,
-)
-from rialto_airflow.snapshot import Snapshot
 from rialto_airflow.publish import publication
-
+from rialto_airflow.schema import harvest as harvest_schema
+from rialto_airflow.schema import rialto as rialto_schema
+from rialto_airflow.schema import reports as reports_schema
+from rialto_airflow.snapshot import Snapshot
```
edsu, line 14: I tried to consolidate these imports to make them a bit easier to identify which of the three sqlalchemy schemas are being used.

test/conftest.py
```diff
@@ -75,25 +62,52 @@ def test_session(test_engine):
...
+    )
+    monkeypatch.setattr(
+        harvest_incremental.openalex, "RIALTO_DB_NAME", "rialto_incremental_test"
+    )
+    monkeypatch.setattr(
+        harvest_incremental.pubmed, "RIALTO_DB_NAME", "rialto_incremental_test"
+    )
+    monkeypatch.setattr(
+        harvest_incremental.wos, "RIALTO_DB_NAME", "rialto_incremental_test"
+    )
```
edsu, line 94: hopefully at some point after the rialto.harvest module has gone away this monkeypatching will no longer be needed...

pyproject.toml
```diff
@@ -43,23 +43,6 @@ addopts = "-v --cov --cov-report=html --cov-report=term --log-level INFO --log-f
...
-    "pyalex",
-    "requests_oauthlib",
-    "sqlalchemy_utils",
-    "jsonpath_ng",
-    "dominate",
-    "dominate.tags",
-    "typing_extensions"
-]
-ignore_missing_imports = true
-
```
edsu, line 62: mypy has been replaced by ty, so this config is unused.

test/conftest.py
lwrubel: That is definitely helpful in making sure we're using the right schema and making things more readable.

lwrubel:
Thanks, this will also make it easier to clean up later when we are no longer using the harvest DAG.

lwrubel (approved):

PR #881, opened 2026-05-06
PubMed incremental harvest

### LINKED ISSUE #847: PubMed incremental harvest
Update the `rialto_airflow.harvest2.pubmed.harvest()` function to use the `Harvest.previous()` method to determine when the last successful harvest *started*. Use that time to fetch records that have been updated since. If no previous Harvest is found then fetch all publications.

edsu:
**What was changed?**

This change adds incremental harvesting for PubMed.

One detail about PubMed harvesting is that if you want to get records
that were modified since a particular date you need to provide both a 
min and max date range.

Closes #847

n/a

### DIFF
--- a/rialto_airflow/harvest_incremental/pubmed.py
+++ b/rialto_airflow/harvest_incremental/pubmed.py
@@ -1,3 +1,4 @@
+import datetime
 import json
 import logging
 import os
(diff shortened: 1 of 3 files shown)


lwrubel (changes requested):
Github is acting flaky and I can't tell if it's saving my comment. Tried to add that when updating/inserting publications, we need to populate the `pubmed_harvested` provenance column by adding something like:

`pubmed_harvested=datetime.datetime.now(datetime.timezone.utc)` 

See `dimensions.py` or other incremental harvesters so far.

lwrubel (approved):

PR #885, opened 2026-05-07
do not bother with funders that are already linked

### LINKED ISSUE #851: Limit funder linking to publications that lack them
Instead of linking *all* publications to funders every time the harvest DAG runs, we should only link publications that currently lack them:

- [ ] `rialto_airflow.funders.link_dim_publications()` should limit to publications that lack a funder
- [ ] `rialto_airflow.funders.link_openalex_publications()` should limit to publications that lack a funder

peetucket:
**What was changed?**

Fixes #851, skip linking if a funder exists already

note: gemini-cli assisted work

### DIFF
--- a/rialto_airflow/funders/linker.py
+++ b/rialto_airflow/funders/linker.py
@@ -39,6 +39,7 @@ def link_dim_publications(database_name: str) -> int:
         stmt = (
             select(Publication)
             .where(Publication.dim_json.is_not(None))
+            .where(~Publication.funders.any())
             .execution_options(yield_per=100)
         )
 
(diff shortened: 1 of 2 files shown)


rialto_airflow/funders/linker.py
```diff
@@ -74,31 +75,32 @@ def link_openalex_publications(database_name: str) -> int:
...
+                for funder in funders:
+                    openalex_funder_id = funder["funder"]
+                    if funder_id := _find_or_create_openalex_funder(
+                        update_session, openalex_funder_id
+                    ):
+                        update_session.execute(
+                            insert(pub_funder_association)
+                            .values(publication_id=pub.id, funder_id=funder_id)
+                            .on_conflict_do_nothing()
+                        )
```
peetucket, line 103: there is nothing different about this code, but gemini noticed it seems to be indented inconsistently from the other similar methods, notably out of scope of the `with get_session` line above.

this indents it to be in the same block as the session

test/funders/test_linker.py
```diff
@@ -49,7 +51,7 @@ def test_link_publications_harvest(test_session, harvest_database_name):
             )
         )
 
-    assert linker.link_publications(harvest_database_name) == 2
+    assert linker.link_publications(harvest_database_name) == 1
```
peetucket, line 54: only one gets linked, the other is skipped

rialto_airflow/funders/linker.py
lwrubel: I wonder how this was working before, out of the session context? This seems right, that the loop through publications should happen within the top-level session context manager.

lwrubel (approved):
Approved pending the code coverage checks of logging.

rialto_airflow/funders/linker.py
peetucket: I wondered the same thing.  Are we sure it actually ever worked?

lwrubel: There are 915,670 rows in the latest harvest database's `pub_funder_association` table, so seems like there are associations being created.

lwrubel: Actually, I didn't restrict that to linkages based on Openalex (which is where the indenting issue was) so let me look further.

lwrubel: It wasn't working :( 
```
rialto_20260503000005=# select count(*) from publication inner join pub_funder_association on publication.id=pub_funder_association.publication_id where publication.openalex_json is null;
 count 
-------
 81727
(1 row)

rialto_20260503000005=# select count(*) from publication inner join pub_funder_association on publication.id=pub_funder_association.publication_id where publication.dim_json is null;
 count 
-------
     0
(1 row)
```

PR #888, opened 2026-05-08
only distill pubs that require it

### LINKED ISSUE #289: Only distill publications that need it
The `distill_publications` step runs on ALL publications, which is desirable when we are harvesting data for the first time. However if the distill step fails, or if we ever start doing incremental updates, it would be helpful if we only ran distill on publications where it hasn't ran before. This DAG Task takes ~8 hours to complete.

One approach to doing this would be to add a distilled timestamp column to the publication table, that is set when the publication is distilled, and to check that it is not NULL when building the list of publications to distill.

peetucket:
**What was changed?**

Fixes #289 

1. New publication column (migation) that records when a publication is distilled (update distill logic)
2. When incremental harvesting, only distill a publication if:
- it has never been distilled (new column is null)
- the last distilled date is before the last time any of the sources for the publication were harvested

note: gemini-cli assisted code and tests


test/test_database.py
```diff
@@ -279,6 +281,37 @@ def test_author_fixture(test_session, author):
...
+    now = datetime.datetime.now(datetime.timezone.utc)
+    pub.distilled_at = now
+    pub.updated_at = now
+    assert pub.needs_distillation() is False
+
+    later = now + datetime.timedelta(seconds=1)
+    pub.updated_at = later
+    assert pub.needs_distillation() is True
+
+
```
peetucket, line 314: is there a better place to put these methods that are on the publication model (defined in the schema file)?

test/harvest_incremental/test_distill.py
```diff
@@ -28,3 +28,35 @@ def test_distill(
         )
         assert pub.academic_council_authored is True
         assert pub.faculty_authored is True
+        assert pub.distilled_at is not None
+
+
+def test_academic_council_authored():
```
lwrubel, line 36: These are new tests added for better coverage?

test/harvest_incremental/test_distill_conditional.py
```diff
@@ -0,0 +1,69 @@
+import datetime
+import time
+from rialto_airflow.harvest_incremental.distill import distill
+from rialto_airflow.schema.rialto import Publication
+
+
+def test_distill_conditional(
```
lwrubel: Was there a reason to put this test in a separate file? It seems like it would be OK to keep this in `test_distill.py` since that file isn't too long.

test/test_database.py
lwrubel: Seems OK to me, but @edsu might have have a suggestion.

lwrubel:
Just checking if this ran locally and was fine? Seems like it would be OK, but would be good to test that way if it hasn't been already.

test/harvest_incremental/test_distill.py
peetucket: that's right, improve codecov coverage

test/harvest_incremental/test_distill_conditional.py
peetucket: no reason, i moved them the existing test_distill.py and it doesn't seem too long

makes sense to keep it together

peetucket:
> Just checking if this ran locally and was fine? Seems like it would be OK, but would be good to test that way if it hasn't been already.

i just ran a local harvest with a limit of 1000 and it completed successfully, trying to think of what else to test...maybe run it again?

test/harvest_incremental/test_distill_conditional.py
edsu: I have a slight preference for keeping the distill tests together unless it is getting too bloated.

test/test_database.py
edsu: I like them as methods on the class.

edsu (approved):

PR #894, opened 2026-05-13
Update publish reports dag to use incremental harvest

### LINKED ISSUE #890: Change publish_to_reports DAG to read from rialto database
The `publish_to_reports` DAG gets a database name from a `Snapshot`. Change it and its tasks to read from the `rialto` database (`RIALTO_DB_NAME` in rialto_airflow.schema.rialto`) instead.

### LINKED ISSUE #931: Prevent publish to reports when harvest is running
Now that we are harvesting to the same database every week we should prevent the `publish_to_reports` DAG from running when there is an active harvest.

The reason for this is that when the harvest is underway the data in the database will be in flux (distilled values changing, unremoved duplicates, etc).

It should be possible to simply search the `harvest` table for the most recently created Harvest and ensure that it has `completed_at` set?

I think we could use a [ShortCircuitOperator](https://airflow.apache.org/docs/apache-airflow-providers-standard/stable/operators/python.html#howto-operator-shortcircuitoperator) to achieve this?

peetucket:
HOLD until incremental harvest is on and working

**What was changed?**

Fixes #890 - updates existing publish DAG for incremental harvests
Fixes #931 - do not allow publish to reports to proceed if a harvest is running
Only merge when ready to switch over to publishing from incremental database.

Note: the fixture data for the incremental database has only one publication, so the tests were altered to match the new fixture data


rialto_airflow/dags/publish_to_reports.py
```diff
@@ -35,19 +35,19 @@ def get_snapshot():
...
 
     @task
     def publish_publications_by_department(snapshot):
-        publication.export_publications_by_department(snapshot)
+        publication.export_publications_by_department(snapshot.database_name)
 
     @task
     def publish_publications_by_author(snapshot):
-        publication.export_publications_by_author(snapshot)
+        publication.export_publications_by_author(snapshot.database_name)
```
peetucket: these are all changes to the existing dag to just pass in the database name instead of the whole snapshot object, this will let us use it functions in `publication` more easily in the new incremental harvest dag

the publish functions are also refactored to use the database name instead of the snapshot object (they didn't need anything other than the the database name anyway)

rialto_airflow/publish/publication.py
```diff
@@ -225,14 +225,14 @@ def export_publications_by_department(snapshot) -> int:
...
-def export_publications_by_author(snapshot) -> int:
+def export_publications_by_author(database_name: str) -> int:
     """
     Export publication and author information to the publications_by_author table.
     """
     logging.info("started writing publications_by_author table")
     count = 0
 
-    with get_session(snapshot.database_name).begin() as select_session:
+    with get_session(database_name).begin() as select_session:
```
peetucket: the refactor of the shared publication functions to remove the dependency on the snapshot object

rialto_airflow/publish/publication.py
lwrubel: For publishing from the incremental harvesting DAG, I think it is a problem to use the non-incremental harvest schema. We would want to use `rialto_airflow.schema.rialto` and its classes.

lwrubel:
Implementing this as side by side DAG but sharing the `publish` code is complicated. The ticket was in the backlog for that reason, although I see I could have explicitly added a "blocked by" to make the reason clear.

peetucket:
> Implementing this as side by side DAG but sharing the `publish` code is complicated. The ticket was in the backlog for that reason, although I see I could have explicitly added a "blocked by" to make the reason clear.

ok, i'll set to draft and mark as hold for now, we can return to it when ready

rialto_airflow/dags/publish_to_reports.py
```diff
@@ -25,41 +24,31 @@
...
 )
 def publish_to_reports():
-    @task()
-    def get_snapshot():
-        snapshot = Snapshot.get_latest(data_dir)
-        if snapshot is None:
-            raise Exception(f"Unable to find completed snapshot in {data_dir}")
-        else:
-            return snapshot
-
```
peetucket, line 35: don't need snapshot anymore or need to pass it the publication functions since we always use the same database name to pull data from now

edsu (approved):
Nice work!

PR #899, opened 2026-05-13
Dimensions fill_in on updated DOIs only

### LINKED ISSUE #895: Dimensions fill_in task should be based on updated/new DOIs
`dimensions.fill_in()` should query only for DOIs for those publications with an `updated_at` date later than the `created_date` of the current `Harvest`. The `harvest_id` should be passed to the `fill_in` tasks. 

This should speed up the `fill_in_dimensions` task.

lwrubel:
**What was changed?**
Fixes #895 

N/A

### DIFF
--- a/rialto_airflow/harvest_incremental/dimensions.py
+++ b/rialto_airflow/harvest_incremental/dimensions.py
@@ -269,14 +269,20 @@ def query_with_retry(q, retry=5):
                 time.sleep(try_count * 10)
 
 
-def fill_in():
+def fill_in(harvest_id: int) -> None:
     """Harvest Dimensions data for DOIs from other publication sources."""
     count = 0
     with get_session(RIALTO_DB_NAME).begin() as select_session:
+        harvest_created_at = (
+            select(Harvest.created_at).where(Harvest.id == harvest_id).scalar_subquery()
+        )
+
         stmt = (
             select(Publication.doi)
             .where(Publication.doi.is_not(None))
             .where(Publication.dim_json.is_(None))
+            .where(Publication.updated_at.is_not(None))
+            .where(Publication.updated_at > harvest_created_at)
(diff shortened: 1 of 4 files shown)


test/conftest.py
```diff
@@ -259,6 +261,25 @@ def test_incremental_session(test_incremental_engine):
...
+        with test_incremental_session.begin() as session:
+            harvest = rialto_schema.Harvest(
+                created_at=created_at,
+                finished_at=finished_at,
+            )
+            session.add(harvest)
+            session.flush()
+            return harvest.id
+
+    return _create
```
lwrubel, line 280: Copilot suggested scoping this return a harvest_id, rather than a Harvest object. I think this works since the Harvest id is what is getting passed around in many tests.

rialto_airflow/harvest_incremental/dimensions.py
```diff
@@ -269,14 +269,20 @@ def query_with_retry(q, retry=5):
...
     with get_session(RIALTO_DB_NAME).begin() as select_session:
+        harvest_created_at = (
+            select(Harvest.created_at).where(Harvest.id == harvest_id).scalar_subquery()
+        )
+
         stmt = (
             select(Publication.doi)
             .where(Publication.doi.is_not(None))
             .where(Publication.dim_json.is_(None))
+            .where(Publication.updated_at.is_not(None))
```
edsu, line 284: I don't think updated_at will ever be NULL (since it should be set on insert), and `updated_at > harvest_created_at` will evaluate to False if it was. But it's not doing any harm to check for it not being NULL.

test/conftest.py
edsu: Can't the fixture simply return the Harvest object or the Harvest.id instead of a function that returns one of them?

edsu (approved):

PR #905, opened 2026-05-13
use edat and relative dates for pubmed incremental harvests

### LINKED ISSUE #904: Change pubmed incremental query params to use edat instead of mdat
Instead of using `mdat` (modified date), `_pubmed_search_api` should query by `edat` (Entrez Date) as sul_pub does: 

https://github.com/sul-dlss/sul_pub/blob/main/lib/pubmed/query_author.rb#L37-L41

To provide the date range, we can use either a `reldate` that corresponds to a number of days or a `mindate` with `maxdate`. Sul_pub uses a fixed `reldate` so maybe it makes the most sense to use the same param (`reldate`). WOS already has logic to calculate the number of dates, so that could be followed:

https://github.com/sul-dlss/rialto-airflow/blob/main/rialto_airflow/harvest_incremental/wos.py#L169-L178

Ideally we could construct a query that could get publications that had either a `mdat` or `edat` in the relevant timeframe, but it's not clear to me that is possible with the API.

peetucket:
**What was changed?**

Fixes #904.  Uses the same params for Pubmed incremental harvesting as we do in sul-pub, since we know it works there.  Refactor out the calculation for "days since" since we now use it in two places.  

note: gemini-cli assisted code.

### DIFF
--- a/rialto_airflow/harvest_incremental/pubmed.py
+++ b/rialto_airflow/harvest_incremental/pubmed.py
@@ -23,7 +23,7 @@ from rialto_airflow.schema.rialto import (
     RIALTO_DB_NAME,
     Harvest,
 )
-from rialto_airflow.utils import normalize_doi, normalize_pmid
+from rialto_airflow.utils import days_since, normalize_doi, normalize_pmid
 
 
 PUBMED_KEY = os.environ.get("AIRFLOW_VAR_PUBMED_KEY")
@@ -279,9 +279,13 @@ def _pubmed_search_api(
     params = {**BASE_PARAMS, "retmode": "json", "term": query}
 
     if previous_harvest:
-        params["datetype"] = "mdat"
-        params["mindate"] = previous_harvest.created_at.strftime("%Y/%m/%d")
-        params["maxdate"] = datetime.date.today().strftime("%Y/%m/%d")
+        params["datetype"] = "edat"
+        # Calculate number of days since the start of the previous finished harvest
+        number_of_days = days_since(previous_harvest.created_at)
+        if number_of_days <= 0:
+            logging.info("Zero days since last harvest, nothing to do.")
+            return []
+        params["reldate"] = number_of_days
 
     logging.debug(f"searching pubmed with {params}")
 
--- a/rialto_airflow/harvest_incremental/wos.py
+++ b/rialto_airflow/harvest_incremental/wos.py
@@ -21,6 +21,7 @@ from rialto_airflow.schema.rialto import (
     Harvest,
 )
 from rialto_airflow.utils import (
+    days_since,
     normalize_doi,
     normalize_pmid,
     normalize_wos_id,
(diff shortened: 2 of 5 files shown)


rialto_airflow/harvest_incremental/wos.py
```diff
@@ -169,10 +170,7 @@ def publications_from_orcid(orcid, harvest_date=None) -> Generator[dict, None, N
     if harvest_date is not None:
         # construct a date limiter that is relative to the current date, since WOK does not allow use of the modifiedTimeSpan limiter
         # date limiter should use the number of days since the start of the previous finished harvest + "D", e.g. "7D"
-        # Although harvest_date will have been saved as a UTC datetime, to avoid a TypeError making sure it's timezone-aware.
-        harvest_date = harvest_date.astimezone(datetime.timezone.utc)
-        current_date = datetime.datetime.now(datetime.timezone.utc)
-        number_of_days = (current_date - harvest_date).days
+        number_of_days = days_since(harvest_date)
```
peetucket, line 173: the new `days_since` function now handles the conversion to UTC, no need to do it here first

rialto_airflow/harvest_incremental/pubmed.py
```diff
@@ -279,9 +279,13 @@ def _pubmed_search_api(
...
-        params["datetype"] = "mdat"
-        params["mindate"] = previous_harvest.created_at.strftime("%Y/%m/%d")
-        params["maxdate"] = datetime.date.today().strftime("%Y/%m/%d")
+        params["datetype"] = "edat"
+        # Calculate number of days since the start of the previous finished harvest
+        number_of_days = days_since(previous_harvest.created_at)
+        if number_of_days <= 0:
+            logging.info("Zero days since last harvest, nothing to do.")
+            return []
+        params["reldate"] = number_of_days
```
peetucket, line 288: if 0 days difference, this does nothing which makes sense to me, but should it instead do everything (by leaving off the reldate param entirely)?  this is what wos seems to do below but i'm not sure why if we already check to see if there was a previous harvest

test/harvest_incremental/test_pubmed.py
```diff
@@ -348,38 +349,54 @@ def test_incremental_harvest(
...
+        pubmed.harvest()
+
+        assert requests_mock.called
+        assert len(requests_mock.request_history) == 1
+        req = requests_mock.request_history[0]
+        assert req.qs["term"] == ["0000-0000-0000-0001[auid]"]
+        assert req.qs["datetype"] == ["edat"]
+        assert req.qs["reldate"] == ["15"]
+        assert "mindate" not in req.qs
+        assert "maxdate" not in req.qs
```
peetucket, line 399: most of this test didn't really change, just some indentation changes due to the new mocking

rialto_airflow/harvest_incremental/pubmed.py
edsu: I agree. I would expect it to do nothing more than I would expect it to harvest everything.

lwrubel: @peetucket and I discussed this and he'll ticket and pick up the adjustment to wos to handle when there are zero days, as well as another issue with the `loadTImeSpan `we came across.

lwrubel (approved):

PR #909, opened 2026-05-14
use corret loadTimeSpan parameter for WOS API calls

### LINKED ISSUE #908: Adjust WoS timespan parameter for incremental harvests
See #905 for shared method
 
1. Currently the WoS API uses an "x" Days since param.  But WoS requires:
- 0-6 is days
- 1-52 is weeks
- 1-12 is years

using D, W, or Y

```
Load time span (otherwise described as symbolic time span) defines a range of load dates. The load date is the date a record was added to the database. If load date is specified, the publishTimeSpan, createdTimeSpan and, modifiedTimeSpan parameter must be omitted. If both publishTimeSpan, loadTimeSpan, createdTimeSpan, and modifyTimeSpan are omitted, the maximum publication date will be inferred from the editions data. Any of D/W/M/Y prefixed with a number where D-Day, M-Month, W-Week, Y-Year allowed. Acceptable value range for Day(0-6), Week(1-52), Month(1-12) and Year(0-10), ex: 5D,30W,10M,8Y.
```

2. Is the number of days <= 0, return an empty array

peetucket:
**What was changed?**

Fixes #908:

1. WoS API timespan requires weeks or years if we get too big a number of days
2. If we end up with 0 day timespan, it should be a noop instead of running a full harvest

### DIFF
--- a/rialto_airflow/harvest_incremental/wos.py
+++ b/rialto_airflow/harvest_incremental/wos.py
@@ -157,6 +157,30 @@ def fill_in(harvest_id: int) -> None:
     logging.info(f"filled in {count} publications")
 
 
+def format_wos_timespan(days: int) -> str:
+    """
+    Format the number of days into a string suitable for the WoS API loadTimeSpan parameter.
+    If the number of days is more than 6, it must be expressed in weeks (W) or years (Y).
+    - <= 6 days: use days (D)
+    - > 6 days and <= 52 weeks: use weeks (W)
+    - > 52 weeks: use years (Y)
+    Note: there is a month parameter (M) but we don't need it since we can go up to 52 weeks
+    """
+    if days <= 6:
+        return f"{days}D"
+
+    weeks = (
+        (days + 6) // 7
+    )  # drop any remainder since the API expects whole weeks, but round up to the next week
+    if weeks <= 52:
+        return f"{weeks}W"
+
+    years = (
+        (days + 364) // 365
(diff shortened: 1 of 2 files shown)


test/harvest_incremental/test_wos.py
```diff
@@ -14,6 +14,22 @@
...
+    class FixedDatetime(datetime.datetime):
+        @classmethod
+        def now(cls, tz=None):
+            fixed_now = cls(2026, 5, 1, 0, 0, 0, tzinfo=datetime.timezone.utc)
+            return fixed_now.astimezone(tz)
+
+    monkeypatch.setattr(wos.datetime, "datetime", FixedDatetime)
+    return FixedDatetime
+
+
```
peetucket, line 32: this was refactored up here since it was being used multiple times now in the new tests too

rialto_airflow/harvest_incremental/wos.py
```diff
@@ -157,6 +157,30 @@ def fill_in(harvest_id: int) -> None:
...
+    if days <= 6:
+        return f"{days}D"
+
+    weeks = (
+        (days + 6) // 7
+    )  # drop any remainder since the API expects whole weeks, but round up to the next week
+    if weeks <= 52:
+        return f"{weeks}W"
+
+    years = (
```
lwrubel, line 178: Hopefully we're running this more frequently than annually, but this means we will be prepared if we do some refresh that goes farther back.

lwrubel (approved):

PR #910, opened 2026-05-15
Remove openalex only pubs from reports

edsu:
**What was changed?**

In order to minimize the impact of bad OpenAlex metadata we are going to not visualize publications that have only come from OpenAlex. The idea is that this is a temporary fix, until OpenAlex metadata improves.

It looks like the changes are pretty simple. But we will need to be careful when querying the primary database in ad-hoc reporting.

One little gotcha here, is that all these tests still target the `harvest` schema. So we will need to update #890, or this PR if #890 is merged before, in order to use the `rialto` database.

Closes #987


peetucket (commented):
Makes sense to me.

lwrubel (approved):
We talked it through with @jacobthill today and are good to go with this being merged and deployed.

lwrubel (approved):

PR #914, opened 2026-05-21
info about Dimensions API key

peetucket:
**What was changed?**

Add information to readme about where Dimensions API key comes from.  Info placed in vault already


lwrubel (approved):
Thanks for doing this!

PR #921, opened 2026-05-28
Optionally do full harvest

### LINKED ISSUE #915: Add a mechanism for triggering a full reharvest
After some analysis performed as part of #853 we decided to add the ability to trigger full reharvests in order to refresh metadata, the so called "brute force" approach to updating.

The idea is that we would keep the existing `rialto` database, and reharvest all the publications by ORCID from all platforms, and let the deduplication task clean up any duplicates (publications without DOI) that are persisted. The distill step will naturally run for all updated publications.

- [ ] Add a `full` boolean value to the `Harvest` table, which defaults to `False`. This will let us have a history of when full reharvests have been performed.
- [ ] Update the `setup` task to create the Harvest with `full` set to `True` when `AIRFLOW_VAR_HARVEST_FULL` is set in the environment.
- [ ] Update `Harvest.get_previous()` to return `None` when the `full` property is `True`.
- [ ] Add a new task `reset_authorship_links` that will run after the `setup` task to remove all links between authors and publications when AIRFLOW_VAR_HARVEST_FULL is True. This is needed because we want to pick up corrected authorship claims (especially for OpenAlex when they are fixed).
- [ ] Add a new task `remove_orphaned_publications` to the `post_process` task group to remove any dangling publications (publications without authors). The removed publication DOI as well as what platform(s) the metadata came from should be logged. This can safely run whether or not we are running a full harvest.

edsu:
**What was changed?**

When users manually trigger the `harvest_incremental` DAG in the Airflow UI they will be given the option to perform a "Full harvest". This means that rather than using the `started_at` of the previous harvest to limit the publications that are retrieved, no limit is used.

Screenshot 2026-05-28 at 1 16 19 PM

The way it works is:

- Add an `is_full` property to the Harvest model, and set it to True when the user has opted for a full harvest. The default is False.
- Pass the `harvest_id` to all the harvest functions
- Harvest functions use harvest's `get_previous()` method (previously a class method) to determine the previous harvest, and which returns None when `harvest.is_full is True`.
- When performing a full harvest all the publication/author links are removed.
- As a final step in the post_process task group the `remove_orphan_pubs` will remove any publications that are not linked to a Stanford author. We might expect these to accumulate when we do a full harvest after platforms fix links between authors and publications.

The logic for `Harvest.get_previous()` changed as part of this commit. Previously it returned the most recently completed Harvest in order of it's `id`. While this worked ok, it was changed to use the `Harvests.created_at` value, to find the closest preceding Harvest that completed, which is now possible since `get_previous()` is now instance method instead of a class method.

I did add a [wiki page](https://github.com/sul-dlss/rialto-airflow/wiki/Incremental-Harvesting) for how to run the full harvest. I also created a separate ticket #922 to update the README.md once everything is complete and the harvest DAG has been removed. 

Fixes #915

n/a


alembic_rialto/versions/550b4eef1eeb_add_harvest_full_column.py
edsu, line 1: Track whether a harvest is a "full harvest" or not.

rialto_airflow/dags/harvest_incremental.py
```diff
@@ -49,14 +49,27 @@
...
     catchup=False,
     default_args=default_args(),
+    params={
+        "full_harvest": Param(
+            False,
+            type="boolean",
+            title="Run a full harvest",
+            description="Collect publications from data sources irrespective of when they were added/updated. This will effectively update all the platform metadata.",
+        )
+    },
```
edsu, line 59: This allows users to specify that a harvest is a "full harvest" when manually starting the workflow.

rialto_airflow/dags/harvest_incremental.py
```diff
@@ -65,41 +78,42 @@ def load_authors(harvest_id):
         Load the authors data from the authors CSV into the database.
         """
         authors.load_authors_table(data_dir)
+        authors.clear_pub_author_links(harvest_id)
```
edsu, line 83: This is a no-op when the harvest isn't a full harvest. when it is a full harvest all the publication/author links will be removed.

rialto_airflow/dags/harvest_incremental.py
```diff
@@ -65,41 +78,42 @@ def load_authors(harvest_id):
...
 
     @task()
     def dimensions_harvest(harvest_id):
         """
         Fetch the data by ORCID from Dimensions.
         """
-        dimensions.harvest(limit=harvest_limit)
+        dimensions.harvest(harvest_id, limit=harvest_limit)
```
edsu, line 90: There are a lot of changes like these where the ID for the Harvest is passed into the harvesting function. This allows us to look up the harvest previous to it when necessary.

rialto_airflow/dags/harvest_incremental.py
```diff
@@ -175,18 +189,24 @@ def link_funders(harvest_id):
         """
         funders.link_publications(RIALTO_DB_NAME)
 
+    @task()
+    def remove_orphan_pubs(harvest_id):
+        """
+        Remove publications that have no linked authors.
+        """
+        deduplicate.remove_orphan_publications()
+
```
edsu, line 200: This step will prune away any publications that are not associated with a Stanford author. full harvests can cause these to appear when a platform has corrected their authorship data (e.g. OpenAlex).

rialto_airflow/harvest_incremental/pubmed.py
```diff
@@ -58,15 +58,16 @@
...
     """
     Walk through all the Author ORCIDs and generate publications for them from pubmed.
     """
     pub_count = 0
     author_count = 0
     stop = False
 
-    previous_harvest = Harvest.get_previous()
+    harvest = Harvest.get_by_id(harvest_id)
+    previous_harvest = harvest.get_previous()
```
edsu, line 70: get_previous() is now an instance method, that returns the harvest that is directly previous to the one being worked on.

test/harvest_incremental/test_dimensions.py
edsu, line 1: Some adjustments to the harvest dates was necessary because the harvest's created_at is used to determine the previous harvest now, instead of the primary key.

test/harvest_incremental/test_harvest.py
```diff
@@ -68,16 +116,11 @@ def test_complete_sets_finished_at(test_incremental_session, mock_rialto_db_name
...
-    assert Harvest.get_by_id(999999) is None
-
-
-def test_get_previous_returns_none_when_no_finished_harvests(
+def test_get_by_id_raises_for_missing_harvest(
     test_incremental_session, mock_rialto_db_name
 ):
-    assert Harvest.get_previous() is None
+    with pytest.raises(ValueError, match="999999"):
+        Harvest.get_by_id(999999)
```
edsu, line 123: get_by_id() now raises an exception instead of returning None, which makes it easier for type checking.

test/harvest_incremental/test_pubmed.py
```diff
@@ -386,36 +388,37 @@ def test_incremental_harvest(
...
+        # harvest from Pubmed (with mocked datetime.now)
+        pubmed.harvest(active_harvest_id)
 
         assert requests_mock.called
         assert len(requests_mock.request_history) == 1
         req = requests_mock.request_history[0]
         assert req.qs["term"] == ["0000-0000-0000-0001[auid]"]
         assert req.qs["datetype"] == ["edat"]
-        assert req.qs["reldate"] == ["15"]
+        assert req.qs["reldate"] == ["22"]
```
edsu, line 399: This needed to change because of changes to the Harvest dates that are created in the test.

test/harvest/test_dimensions.py
```diff
@@ -14,6 +14,7 @@
 dotenv.load_dotenv()
```
edsu, line 16: I skipped live tests of the Dimensions API until our keys are working again. There is ticket #923 as a reminder to re-enable them.

peetucket:
The code coverage drop is because the MAIS tests are no longer being run (because we removed the API key from github secrets for security reasons).  I see the same thing on my PR.   see https://app.codecov.io/gh/sul-dlss/rialto-airflow/pull/921/indirect-changes

I'm not sure if after merging one of our PRs it will realize the code coverage is lower on main overall and stop complaining about the diff drop, like it is here.

rialto_airflow/dags/harvest_incremental.py
```diff
@@ -175,18 +189,24 @@ def link_funders(harvest_id):
...
         link = link_funders(harvest_id)
-        dedupe >> [distill, link]
+        orphans = remove_orphan_pubs(harvest_id)
+        dedupe >> [distill, link] >> orphans
 
     @task()
     def complete(harvest_id):
         harvest = Harvest.get_by_id(harvest_id)
-        if harvest is None:
-            raise ValueError(f"Harvest {harvest_id} not found")
```
peetucket, line 189: just curious if there was a reason this guard was removed

rialto_airflow/harvest_incremental/pubmed.py
peetucket: I feel like a small refactor could be to create a new class level method that does this in one call since it is repeated in many spots (i.e. instead of instantiating the Harvest object first each time).

peetucket: Oh i see, the previous_harvest object can be updated later on the method

edsu: Yeah, the Harvest object is needed in order to determine what the previous harvest is. But. I guess there could be a class method that takes a harvest_id and returns the previous harvest object?

rialto_airflow/dags/harvest_incremental.py
edsu: `get_by_id()` will now throw an exception if the harvest ID is not found in the database, which cause it to fail pretty loudly and stop the workflow (by design)

peetucket (approved):

rialto_airflow/harvest_incremental/dimensions.py
```diff
@@ -270,21 +274,32 @@ def query_with_retry(q, retry=5):
...
             .execution_options(yield_per=100)
         )
 
+        if harvest.is_full:
+            stmt = stmt.where(Publication.dim_harvested.is_(None))
+        else:
+            stmt = stmt.where(Publication.updated_at >= harvest.created_at).where(
+                Publication.dim_json.is_(None)
+            )
+
```
edsu, line 302: There are several bits of logic in other fill-in functions similar to the above to ensure we fill-in all metadata that hasn't just been harvested from Dimensions.

PR #928, opened 2026-06-02
Restore cap_profile_id duplicate handling

edsu:
I think we need to put back some of the code we removed in rialto_airflow author loading to handle the cap_profile_id constraint (see #927). Here's a short scenario to demonstrate why:

1. rialto-orgs gets a user from Profiles

```csv
sunet,cap_profile_id
jsmith,123
```

2. rialto-airflow loads

```csv
sunet,cap_profile_id
jsmith,123
```

3. rialto-orgs gets a new user from Profiles

```csv
sunet,cap_profile_id
jsmith,123
jsmith2,456
```

4. rialto-airflow loads

```csv
sunet,cap_profile_id
jsmith,123
jsmith2,456
```

5. rialto-orgs loads merged data from Profiles

```csv
sunet,cap_profile_id
jsmith2,123
```

6. rialto-airflow load fails

It attempts to upsert `jsmith2` but gets a constraint violation for `cap_profile_id` because a row already exists with `cap_profile_id=123` for `sunet=jsmith`.

### DIFF
--- a/rialto_airflow/harvest_incremental/authors.py
+++ b/rialto_airflow/harvest_incremental/authors.py
@@ -2,6 +2,7 @@ import csv
 import logging
 
 from psycopg2 import Error as Psycopg2Error
+from sqlalchemy import update
 from sqlalchemy.exc import IntegrityError
 from sqlalchemy.orm import Session
 
@@ -52,7 +53,8 @@ def load_authors_table(data_dir) -> None:
                     )
 
                 elif constraint == "author_cap_profile_id_key":
-                    raise e
+                    handle_duplicate_cap_profile_id(session, row_dict)
+                    updated_authors += 1
 
         logging.info(
             f"processed={processed_authors} new={new_authors} updated={updated_authors} ignored={ignored_authors}"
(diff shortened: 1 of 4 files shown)


rialto_airflow/harvest_incremental/authors.py
```diff
@@ -105,6 +107,33 @@ def upsert_author(session: Session, row_dict: dict) -> str:
     return "updated"
 
 
+def handle_duplicate_cap_profile_id(session: Session, row_dict: dict) -> None:
```
edsu, line 110: This is a bit simpler than it was before since we don't need to try to do different things if the user in the database with the cap_profile_id is active or not. We can just rely on rialto-orgs to tell us the correct information now that it has a unique constraint on cap_profile_id.

test/harvest_incremental/test_authors.py
```diff
@@ -191,38 +190,53 @@ def test_load_dupe_orcid(
...
+    # the old sunet should be gone but the publications should be
+    # preserved on the new sunet
     with test_incremental_session.begin() as session:
         assert session.query(Author).count() == 1
+        author = session.query(Author).where(Author.sunet == "lelands2").first()
+        assert author, "found new author"
+        assert author.status is True, "they are active"
+        assert len(author.publications) == 1, (
+            "removal of old author preserved their publications"
+        )
```
edsu, line 239: We didn't need to bring back all the unit tests we had before, since the conditional logic about whether the user is active or not is no longer necessary, and is not being reinstated.

test/harvest/test_openalex.py
edsu, line 1: For some reason OpenAlex seems to be raising an exception that pyalex is not longer wrapping, so this test needed to be updated. OpenAlex is still not liking DOIs with commas in them, so no change there.

peetucket (approved):

PR #932, opened 2026-06-05
Remove the harvest DAG

### LINKED ISSUE #856: Remove harvest DAG
Once the `harvest_incremental` DAG is operational we can remove the `harvest` DAG and its dependent code in `rialto_airflow.harvest` and `rialto_airflow.snapshot`. There also are probably going to be some functions that can be removed from `rialto_airflow.database` that are no longer being used.

### LINKED ISSUE #891: Drop the snapshots task from the cleanup DAG
Once we're no longer running the `harvest` DAG, remove the `snapshots` task from the `cleanup` DAG. It was deleting old databases and snapshot folders with JSON data.

edsu:
**What was changed?**

Remove the harvest DAG all its associated code and tests. Now we just have the
harvest_incremental DAG, which maybe could be renamed to harvest at some point? But maybe not as part of this PR since it would add a layer of confusion.

The commit also attempts to simplify the use of database names in our test fixtures, now that we just have the two databases that we need to interact with.

Apologies, this PR is virtually unreviewable. I have run the code with a limit of 50 in my development environment. We can deploy to stage for a test, but it will take some time with a limit of 50k...

Closes #856
Closes #891 

(this PR is evil)


peetucket (approved):

compose.prod.yaml
```diff
@@ -57,7 +57,6 @@ x-airflow-common:
     AIRFLOW_UID: 503
     AIRFLOW_GID: 0
     AIRFLOW__CORE__EXECUTOR: CeleryExecutor
-    AIRFLOW__CORE__XCOM_BACKEND: rialto_airflow.xcom.RialtoXCom
     AIRFLOW__DATABASE__SQL_ALCHEMY_CONN: "postgresql+psycopg2://${DATABASE_USERNAME}:${DATABASE_PASSWORD}@${DATABASE_HOSTNAME}/rialto-airflow"
```
peetucket, line 60: not needed anymore but unrelated presumably?

edsu: This was used by the incremental harvest DAG to allow the Snapshot to be persisted to the XCom. Now we are passing the harvest_id between DAG tasks which doesn't require any special serialization to JSON.

PR #933, opened 2026-06-05
Use _harvested datetimes to identify OpenAlex only

### LINKED ISSUE #930: Use *_harvested columns for filtering out OpenAlex
Once #894 is merged we will want to update the logic to filter out OpenAlex only publications by using the our new columns that track whether we harvested the publication from a given platform by ORCID. This will be superior to simply using the presence/absence of the platform JSON fields since those could have been filled in after the fact in cases where we got the DOI from OpenAlex.

The columns are:

- openalex_harvested
- wos_harvested
- dim_harvested
- sulpub_harvested
- pubmed_harvested

There are a few different places in the publish to reports process where the condition needs to be switched to use these columns, for example:

https://github.com/sul-dlss/rialto-airflow/blob/eb61e1cfd878a23d96e025df8f599d87107973a0/rialto_airflow/publish/publication.py#L78-L82

edsu:
**What was changed?**

We want to limit publications we write to the reports database (which is
visualized by Tableau), but currently we are only using the presence of
the JSON metadata as an indicator of whether it only came from OpenAlex.
The problem with this is that we "fill-in" metadata by DOI after the
fact. If we use the `*_harvested` columns instead we will be guaranteed
to filter out everything that was harvested from OpenAlex, irrespective
of whether we were able to fill in the metadata from another by platform
later.

Closes #930


test/publish/test_publication.py
```diff
@@ -111,15 +111,19 @@ def test_limit_openalex_only(
...
         )
-        pub.sulpub_json = None
-        pub.dim_json = None
-        pub.wos_json = None
-        pub.pubmed_json = None
+        pub.openalex_harvested = datetime.datetime.now()
+        pub.sulpub_harvested = None
+        pub.dim_harvested = None
+        pub.wos_harvested = None
+        pub.pubmed_harvested = None
```
edsu, line 126: Here we are setting up the publication so that it looks like it only was harvested by OpenAlex.

test/conftest.py
```diff
@@ -452,17 +452,24 @@ def dataset_incremental(
...
+            dim_harvested=now,
             openalex_json=openalex_json,
+            openalex_harvested=now,
             wos_json=wos_json,
+            wos_harvested=now,
             sulpub_json=sulpub_json,
+            sulpub_harvested=now,
             pubmed_json=pubmed_json,
+            pubmed_harvested=now,
             crossref_json=crossref_json,
```
edsu, line 473: The fixture data needs to set the `*_harvested` colums too.

peetucket (approved):

